PR #2, opened 2017-03-17
Develop

mknblch:


jexp:
In general please try to squash commits into meaningful units with a good message.

PR #32, opened 2017-03-24
Replace PrimitiveLongIntMap with (hppc) LongIntMap

### LINKED ISSUE #13: Remove nondeterminism in HeavyGraphParallelLoadingTest
The HeavyGraphParallelLoadingTest occasionally fails with AIOOBEs thrown by the GraphFactory. Tests should always be deterministic.

knutwalker:
There is a race between [packing](https://github.com/neo4j/neo4j/blob/89b7bf20d72f757945f8c32988e4e4c1c83ce013/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/hopscotch/LongKeyIntValueTable.java#L65) and [unpacking](https://github.com/neo4j/neo4j/blob/89b7bf20d72f757945f8c32988e4e4c1c83ce013/community/primitive-collections/src/main/java/org/neo4j/collection/primitive/hopscotch/PrimitiveLongIntHashMap.java#L81) the found value, when multiple threads are accessing the IdMap. The LongIntHashMap, while not being safe to write, is safe to read concurrently.

Fixes #13


knutwalker:
#33 as a follow up and/or consequence.

PR #35, opened 2017-03-24
Improve GraphLoader API

knutwalker:
Note: PR Check might fail due to #13 which is fixed in #32, but the code is not included here.


knutwalker:
We can merge #32 first and I'll rebase this one to get a stable build.

Issue #37, opened 2017-03-27
Fix weight mapping in heavy graph


mknblch:
org.neo4j.graphalgo.core.heavyweight.HeavyGraphTest is currently ignored due to failing tests (weighted iterator / weighted forEach). Fix weight mapping and unignore test.

mknblch:
duplicate

Issue #6, opened 2017-03-24
Custom default weight


knutwalker:
The WeightMapping assumes a default weight of `0.0`. The number should be configurable per algorithm. Since the defaults aren't stored, this would allow algorithms to reduce memory usage by not storing default weights if they require a different default from `0.0`

PR #43, opened 2017-03-27
extracted iface from WeightMap

### LINKED ISSUE #22: Turn WeightMapping into interface with two cases
Currently we have conditional checks on any get the see if we have a mapping at all or if we just have to return the default value.
We should turn WeightMapping into an interface with at most two implementations.
One is the current implementation and the second one always returns a default value and never stores.

mknblch:
extracted iface from WeightMap, added nullobj. impl, adapted code. should fix #22 & #37


mknblch:
also unignored HeavyGraphTest

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
```diff
@@ -2,48 +2,36 @@
...
+import org.neo4j.graphalgo.api.WeightMapping;
 
 /**
  * single weight cache
  */
-public final class WeightMapping {
+public final class WeightMap implements WeightMapping {
 
     private final int capacity;
     private LongDoubleMap weights;
```
jexp: for efficiency reasons we might want to look into one that's backed by an int -> int map.
But we can file that as an issue for future development

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
```diff
@@ -2,48 +2,36 @@
...
 
-    public WeightMapping(final int capacity) {
+    public WeightMap(final int capacity) {
         this.capacity = capacity;
     }
 
-    public WeightMapping(final int capacity, LongDoubleMap weights) {
+    public WeightMap(final int capacity, LongDoubleMap weights) {
```
jexp: should this be package private?

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
```diff
@@ -2,48 +2,36 @@
 
 import com.carrotsearch.hppc.LongDoubleHashMap;
 import com.carrotsearch.hppc.LongDoubleMap;
+import org.neo4j.graphalgo.api.WeightMapping;
 
 /**
  * single weight cache
  */
-public final class WeightMapping {
+public final class WeightMap implements WeightMapping {
```
jexp: this should also handle default values? i.e not store them and return them for not-found entries

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
```diff
@@ -2,48 +2,36 @@
...
     /**
      * return the weight for id or 0.0 if unknown
      */
+    @Override
     public double get(long id) {
-        if (weights != null) {
-            return weights.get(id);
-        }
-        return 0d;
+        return weights.get(id);
```
jexp: this will return -1 for not-found entries.

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
```diff
@@ -69,7 +57,7 @@ private double extractValue(Object value) {
 
     private void put(long key, double value) {
         if (weights == null) {
-            weights = new LongDoubleHashMap(capacity);
+            weights = new LongDoubleHashMap(capacity); // TODO move to ctor
```
jexp: can we handle this todo here?
otherwise we will also potentially get NPEs in `get`

tests/src/main/java/org/neo4j/graphalgo/core/WeightMappingSerialization.java
```diff
@@ -54,9 +55,9 @@ public static WeightMapping read(DataInput in) throws IOException {
                         in.readVLong(),
                         Double.longBitsToDouble(in.readLong()));
             }
-            return new WeightMapping(capacity, weights);
+            return new WeightMap(capacity, weights);
         }
-        return new WeightMapping(capacity);
+        return new NullWeightMap(0.0); // TODO supply default value
```
jexp: can we turn these default-values todo's into an issue so we don't forget?

knutwalker: Defaults are covered in #6

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
knutwalker: The HPPC maps return 0 as the default, with #6 we have to change this anyway to return the custom default.

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
mknblch: This is what the NullWeightMap does. It returns the given defaultValue for every id.

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
mknblch: I tried not to touch the current implementation until we find a good way to implement this (thread safe adding of values)

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
mknblch: its currently used in 2 different packeges (heavyweight and lightweight)

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
```diff
@@ -2,48 +2,40 @@
...
-    /**
-     * add weight
-     */
-    public void add(long id, double weight) {
-        if (weight != 0) {
-            put(id, weight);
-        }
+    @Override
+    public void set(long id, Object value) {
+        put(id, extractValue(value));
```
knutwalker: Could we add the `if (weight != 0)` (or rather `if (weight != defaultWeight)`) check from the old add method, so that we don't store the default weights?

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
jexp: Ok, sure.

core/src/main/java/org/neo4j/graphalgo/core/WeightMap.java
jexp: Yes!

tests/src/main/java/org/neo4j/graphalgo/core/WeightMappingSerialization.java
```diff
@@ -5,6 +5,7 @@
 import com.carrotsearch.hppc.cursors.LongDoubleCursor;
 import org.apache.lucene.store.DataInput;
```
jexp: is it correct that we use Lucene's DataInput / DataOutput here?

knutwalker: Yes. It's only for File serialization and we use the vint encoding to reduce file size by quite a bit (2x-3x on the wikipedia graph). In the future, we can copy the vint encoding into the repo and not use lucene's classes directly here.

PR #45, opened 2017-03-28
Default weight

mknblch:
This commit enables the GraphLoader to specify a default for unknown property values. It also tries to ease the use of the GraphFactory Ctor by replacing the current signature with a dedicated DTO which contains the setup options for building the graph.


core/src/main/java/org/neo4j/graphalgo/api/GraphSetup.java
```diff
@@ -0,0 +1,94 @@
...
+    // end label type (not yet implemented).
+    public final String endLabel;
+    // relationtype name. null means any relation.
+    public final String relationshipType;
+    // propertyName. null means NO property (the default value will be used instead).
+    public final String propertyName;
+    // default property is used for weighted relationships if property is not set.
+    public final double propertyDefaultValue;
+    // the executor service for parallel execution. null means single threaded evaluation.
+    public final ExecutorService executor;
```
jexp: Not sure if mixing the Executor with declarative graph information is such a good idea.
We can leave it in for now, but should pull it out later.

core/src/main/java/org/neo4j/graphalgo/core/GraphLoader.java
```diff
@@ -129,23 +128,41 @@ public GraphLoader withAnyRelationshipType() {
...
     }
 
     /**
      * Instructs the loader to not load any weights.
+     * The graph is initialized with the propertyDefaultValue instead.
+     *
+     * @param propertyDefaultValue the default value.
+     * @return itself to enable fluent interface
+     */
+    public GraphLoader withoutWeights(double propertyDefaultValue) {
```
jexp: perhaps a better method name, i.e. `withDefaultWeight`

jexp: Oh perhaps I mixed it up, I thought this was the method to provide a default weight if there is no property on a relationship. That one seems to be missing then?

PR #48, opened 2017-03-28
adapted method name & javadoc to PR comments

mknblch:


core/src/main/java/org/neo4j/graphalgo/core/GraphLoader.java
```diff
@@ -142,21 +141,21 @@ public GraphLoader withWeightsFromProperty(String property, double propertyDefau
...
-     * The graph is initialized with the propertyDefaultValue instead.
+     * Instructs the loader to not load any weights. Instead each weight is set
+     * to propertyDefaultValue.
      *
      * @param propertyDefaultValue the default value.
      * @return itself to enable fluent interface
      */
-    public GraphLoader withoutWeights(double propertyDefaultValue) {
+    public GraphLoader withDefaultWeight(double propertyDefaultValue) {
         this.property = null;
```
jexp: this is confusing and not what I meant.
I think there need to be 2 methods:

1. don't load weights, use the weight provided
2. load weights with the given property, if there is no such property use the given weight as the default (and also use that default to avoid storing weight-values that are equal to the default)

knutwalker: There are methods for this:

1. is https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/307ef268b58d181169a54c266cf7151fb5c75786/core/src/main/java/org/neo4j/graphalgo/core/GraphLoader.java#L150
2. https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/307ef268b58d181169a54c266cf7151fb5c75786/core/src/main/java/org/neo4j/graphalgo/core/GraphLoader.java#L137

This PR just renamed the method for 1

Issue #46, opened 2017-03-28
Restrict GraphView to use only supplied label/relation/property/defaultValue


mknblch:
The GraphView doesn't care about the restrictions given in its constructor. To implement more UnitTests we need a fully working GraphView. 

- implement restriction for label-, relation and property types.

mknblch:
This could be fixed with issue #11

PR #50, opened 2017-03-28
Restrict view #46

mknblch:
The commit adds restriction to label-, relation-, and property-types. The transaction handling has also been changed according to the PR comments. Now every method which uses the neo4j core api is wrapped into its own transaction.


jexp:
A general comment, without looking at the code, it will probably slow down quite a lot to wrap each method. 

Going forward the transaction control should be outside the graph view and the graph view is just participating in the running tx that's started at thread start and finished at thread end.

core/src/main/java/org/neo4j/graphalgo/core/neo4jview/RelationIteratorImpl.java
```diff
@@ -1,27 +1,71 @@
...
+    private final RelationCursor cursor;
+    private final long originalNodeId;
+
+    RelationIteratorImpl(Graph graph, GraphDatabaseAPI api, int sourceNodeId, Direction direction, int relationTypeId) throws EntityNotFoundException {
+        this.graph = graph;
+        this.api = api;
+        transaction = api.beginTx();
+        statement = api.getDependencyResolver()
+                .resolveDependency(ThreadToStatementContextBridge.class)
+                .get();
```
jexp: this works if we assume that the iterators are only called within one thread, at minimum we have to document that in the GraphView

core/src/main/java/org/neo4j/graphalgo/core/neo4jview/GraphView.java
```diff
@@ -10,98 +10,89 @@
...
             }
-        } catch (EntityNotFoundException e) {
-            throw new RuntimeException(e);
-        }
+            return null;
+        });
     }
 
     @Override
     public Iterator<WeightedRelationCursor> weightedRelationIterator(int nodeId, Direction direction) {
```
jexp: document on all iterator methods that their resulting iterators can only be used safely the single, calling thread.

core/src/main/java/org/neo4j/graphalgo/core/neo4jview/WeightedRelationIteratorImpl.java
```diff
@@ -1,40 +1,92 @@
...
+        final RelationshipItem relationshipItem = relationships.get();
+        cursor.targetNodeId = graph.toMappedNodeId(relationshipItem.otherNode(originalNodeId));
+        final Cursor<PropertyItem> propertyCursor = relationshipItem.property(propertyKey);
+        if(!propertyCursor.next()) {
+            cursor.weight = defaultWeight;
+            return cursor;
         }
+
+        final PropertyItem propertyItem = propertyCursor.get();
+        cursor.weight = (double) propertyItem.value();
```
jexp: use ((Number)propertyItem.value()).doubleValue()

knutwalker:
I think we have a thread safe view in another issue, it's not supposed to be safe yet.

core/src/main/java/org/neo4j/graphalgo/core/neo4jview/RelationIteratorImpl.java
mknblch: We think it is very uncommon to use one iterator from several threads. I agree with you to document this.

PR #53, opened 2017-03-30
Add more tests

knutwalker:
also fixes some implementations that failed with the new tests. requires j8u>=65 due to bug in generic inference in older java versions.


core/src/main/java/org/neo4j/graphalgo/core/neo4jview/RelationIteratorImpl.java
```diff
@@ -14,13 +14,11 @@
 import org.neo4j.storageengine.api.RelationshipItem;
 
 import java.io.Closeable;
-import java.io.IOException;
 import java.util.Iterator;
```
jexp: please add a short javadoc note about thread safety

tests/src/test/java/org/neo4j/graphalgo/core/CypherExporter.java
```diff
@@ -24,15 +25,27 @@ static void export(PrintWriter out, GraphDatabaseService db) {
...
             tx.success();
         }
         out.flush();
     }
+
+    private static String rel(Relationship rel) {
+        String props = rel.getAllProperties()
+                .entrySet()
+                .stream()
+                .map(e -> e.getKey() + ":" + e.getValue())
```
jexp: works only for non-strings ? and keys are not quoted with backticks

knutwalker: Currently, we only use this for tests that generate a random graph in cypher so that we can "export" the graph and dump it in case of test failures, if we want to reproduce the test with the graph that has failed. Since we only generate the property `weight` with doubles, we're good here.
I'll add a note to the class that it's not intended to be used by anyone else besides the test.

Issue #39, opened 2017-03-27
First clustering graph algorithm


jexp:
Something that's useful / useable, like label-prop or union-find, I leave the choice up to you.

Reasons:

- we have a practical use-case that would benefit from it
- we want to exercise the graph-API also from different kinds of algorithms

One relevant feature would be to consider the "weight" property in a relationship for the "strength" of the connection to a cluster.

As a simple solution to start with, could be to filter relationships to consider  "weight" as a filter, e.g. only consider relationship-input that exceeds a certain weight at all.

PR #56, opened 2017-03-30
Algorithm #39 UnionFind WIP

mknblch:
This one is still work in progress please dont merge yet. I implemented a disjoint-set-struct which will be needed for some algorithms.


core/src/main/java/org/neo4j/graphalgo/core/utils/DisjointSetStruct.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    /**
+     * find setId of element p.
+     * @param p the element in the set we are looking for
+     * @return an id of the set it belongs to
+     */
+    public int find(int p) {
+        if (parent[p] == p) return p;
```
jexp: should we save repeated array access with local variables?

core/src/main/java/org/neo4j/graphalgo/core/utils/DisjointSetStruct.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final int[] parent;
+    private final int[] depth;
+
+    /**
+     *
+     * @param capacity the capacity (maximum node id)
+     */
+    public DisjointSetStruct(int capacity) {
+        parent = new int[capacity];
```
jexp: can we handle this with less memory usage? e.g. with a bitset?

algo/src/main/java/org/neo4j/graphalgo/impl/UnionFind.java
```diff
@@ -0,0 +1,41 @@
...
+        return struct;
+    }
+
+    // TODO visibility
+    @Override
+    public void accept(int sourceNodeId, int targetNodeId, long relationId) {
+        if (!struct.connected(sourceNodeId, targetNodeId)) {
+            struct.union(sourceNodeId, targetNodeId);
+        } else {
+            System.out.println("cycle");
```
jexp: probably record it somewhere

core/src/main/java/org/neo4j/graphalgo/core/utils/DisjointSetStruct.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * find setId of element p.
+     * @param p the element in the set we are looking for
+     * @return an id of the set it belongs to
+     */
+    public int find(int p) {
+        if (parent[p] == p) return p;
+        // path compression optimization
+        parent[p] = find(parent[p]); // balance tree while traversing
```
jexp: recursion, can we have huge depths in larger graphs? perhaps replace with iteration? (i.e. a loop)

core/src/main/java/org/neo4j/graphalgo/core/utils/DisjointSetStruct.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * join set of p (Sp) with set of q (Sq)
+     *
+     * @param p an item of Sp
+     * @param q an item of Sq
+     */
+    public void union(int p, int q) {
+        final int pSet = find(p);
```
jexp: potential check if p = q?

core/src/main/java/org/neo4j/graphalgo/core/utils/DisjointSetStruct.java
mknblch: You are true. I've seen 2 possible optimizations in the wild. One is path-halving which works nonrecursive but might at the end not be as efficient as this one. The other one is path compression by recursivly shortening the path during traversion. According to wiki we should get an overall complexity of O(a(n)) ( a(..) is Inverse-Ackermann function). Anyways.. I'd like to do some real-world tests and benchmars on both implementations.

algo/src/main/java/org/neo4j/graphalgo/impl/UnionFind.java
mknblch: Sure. This class is still 99% WIP.

core/src/main/java/org/neo4j/graphalgo/core/utils/DisjointSetStruct.java
mknblch: The parent array is more like an index-to-parentIndex map. We can switch to an implementation based on linked lists which could grow on-demand but as usual decrease performance. We might consider having both implementations and some kind of configuration to choose one.

jexp:
Lemme know how far you got.

mknblch:
delayed until api2 refactoring is done

mknblch:
still WIP, please do not merge yet.

algo/src/main/java/org/neo4j/graphalgo/impl/MSTPrim.java
```diff
@@ -28,63 +33,95 @@ public MSTPrim(Graph graph) {
...
+        public MinimumSpanningTree(int expectedElements) {
+            costMap = new LongDoubleScatterMap(expectedElements);
         }
 
-        /**
-         * get the element count in this mst
-         */
-        public int size() {
-            return offset;
+        public void add(int sourceNodeId, int targetNodeId, double cost) {
```
mknblch: We split the value into ints in line 57 and combine them again in 80. Change method signature (combinedSourceTarget:long, cost:double):void

tests/src/test/java/org/neo4j/graphalgo/impl/ThresholdUnionFindTest.java
```diff
@@ -0,0 +1,93 @@
...
+        a = idMapper.toMappedNodeId(n1);
+        b = idMapper.toMappedNodeId(n2);
+        c = idMapper.toMappedNodeId(n3);
+        d = idMapper.toMappedNodeId(n4);
+        e = idMapper.toMappedNodeId(n5);
+        f = idMapper.toMappedNodeId(n6);
+    }
+
+    /**
+     * should split the graph into 2 distinct sets
```
mknblch: 3

mknblch:
can be merged. 

- unionFind clustering algorithm (#39) + several data structures + test + benchmark
- changes the ConsumerIFaces according to #68
- implements Weights as datasource #73

core/src/main/java/org/neo4j/graphalgo/core/utils/RawValues.java
```diff
@@ -35,4 +35,32 @@ public static int getHead(long combinedValue) {
...
+
+    /**
+     * convert property value to double
+     *
+     * @param value the value object
+     * @param propertyDefaultValue default value if property cant be converted
+     * @return double representation of value
+     */
+    public static double extractValue(Object value, double propertyDefaultValue) {
+        if (value instanceof Number) {
```
jexp: probably add an early null check for returning default values.

PR #57, opened 2017-03-30
Algorithm MST Prim

mknblch:
Work in progress, please do not merge yet. Implements minimum weighted spanning tree by prim and adds a long impl. of MinPrioQueue


mknblch:
Might be merged!

algo/src/main/java/org/neo4j/graphalgo/impl/MSTPrim.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public MST compute(int startNode) {
+
+        final LongMinPriorityQueue queue = new LongMinPriorityQueue();
+        final MST mst = new MST(graph.nodeCount());
+        final BitSet visited = new BitSet(graph.nodeCount());
+
+        // initially add all relations from startNode to the priority queue
+        visited.set(startNode);
+        graph.forEachRelation(startNode, Direction.BOTH, (sourceNodeId, targetNodeId, relationId, weight) -> {
```
jexp: direction potentially configurable? do we actually need BOTH?

mknblch: I think we need both directions. If we would follow only outgoing edges the algorithm would produce false results. if we start for example at a node with only incoming connections or if any node in the path has only incoming connections the MST would terminate and the MST would be incomplete.

jexp: I'm just concerned that we're doing double the work, b/c one nodes outgoing relationship is the other nodes incoming one. But I trust you on this.

core/src/main/java/org/neo4j/graphalgo/core/utils/LongMinPriorityQueue.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * @return the least element of the queue in constant time.
+     */
+    public long top() {
+        // We don't need to check size here: if maxSize is 0,
+        // then heap is length 2 array with both entries null.
+        // If size is 0 then heap[1] is already null.
+        return heap[1];
```
jexp: but null is not assignable to Long ? so perhaps use an indicator line Long.MIN_VALUE ?

core/src/main/java/org/neo4j/graphalgo/core/utils/LongMinPriorityQueue.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Removes all entries from the queue, releases all buffers.
+     * The queue can no longer be used afterwards.
+     */
+    public void release() {
+        size = 0;
+        heap = null;
+
+        costs.keys = new long[0];
```
jexp: shouldn't all of that happen in costs.clear() ? or costs.release() ?

core/src/main/java/org/neo4j/graphalgo/core/utils/LongMinPriorityQueue.java
```diff
@@ -0,0 +1,198 @@
...
+            if (k <= size && lessThan(heap[k], heap[j])) {
+                j = k;
+            }
+        }
+        heap[i] = node;            // install saved node
+    }
+
+    private void ensureCapacityForInsert() {
+        if (size >= heap.length) {
+            long[] newHeap = new long[ArrayUtil.oversize(
```
jexp: there is also Arrays.copyOf() which you might want to consider.

core/src/main/java/org/neo4j/graphalgo/core/utils/LongMinPriorityQueue.java
```diff
@@ -0,0 +1,198 @@
...
+            int i = 1;
+
+            @Override
+            public boolean hasNext() {
+                return i <= size;
+            }
+
+            @Override
+            public long next() {
+                if (!hasNext()) {
```
jexp: don't need to check every time, the array itself has an overflow check which will throw the exception

core/src/main/java/org/neo4j/graphalgo/core/utils/RawValues.java
```diff
@@ -0,0 +1,38 @@
...
+        return (int) (combinedValue >> 32);
+    }
+
+    /**
+     * get the tail value
+     * @param combinedValue a value built of 2 ints
+     * @return the least significant 4 bytes as int
+     */
+    public static int getTail(long combinedValue) {
+        return (int) combinedValue;
```
jexp: I think you have to do (combinedValue & 0xFFFFFFFFL) first to null out potential leading set first bits?

core/src/main/java/org/neo4j/graphalgo/core/utils/RawValues.java
```diff
@@ -0,0 +1,38 @@
...
+        return ((long) head << 32) | tail & 0xFFFFFFFFL ;
+    }
+
+    /**
+     * get the head value
+     * @param combinedValue a value built of 2 ints
+     * @return the most significant 4 bytes as int
+     */
+    public static int getHead(long combinedValue) {
+        return (int) (combinedValue >> 32);
```
jexp: might need to use `>>>`

core/src/main/java/org/neo4j/graphalgo/core/utils/RawValues.java
```diff
@@ -0,0 +1,38 @@
+package org.neo4j.graphalgo.core.utils;
+
+/**
+ * TODO: find suitable name or move
+ *
+ * @author mknblch
+ */
+public class RawValues {
```
jexp: do we have tests for edge cases?

PR #60, opened 2017-04-05
Add Shortest Path algorithm

knutwalker:
Implementation from the initial test repository.
Also, added the tests for this one and for PageRank as well. These tests plus one that was previously ignored uncovered some implementation errors, which are fixed as well (most of these are about what to load and not load).
In order to make all tests pass, we have to load the node ids differently. We need a single pass upfront, that fills the id mapping and we build the reverse lookup array in a single step afterwards.

Note: This PR Includes #59 as the new int queue is needed for Dijkstra's algorithm. We should merge #59 first, this will also decrease the volume of this PR.


algo/src/main/java/org/neo4j/graphalgo/impl/ShortestPathDijkstra.java
```diff
@@ -0,0 +1,86 @@
...
+import com.carrotsearch.hppc.IntDoubleScatterMap;
+import com.carrotsearch.hppc.IntIntMap;
+import com.carrotsearch.hppc.IntIntScatterMap;
+import org.neo4j.graphalgo.api.Graph;
+import org.neo4j.graphalgo.core.utils.IntPriorityQueue;
+import org.neo4j.graphalgo.core.utils.SharedIntMinPriorityQueue;
+import org.neo4j.graphdb.Direction;
+import org.neo4j.kernel.impl.util.collection.SimpleBitSet;
+
+public class ShortestPathDijkstra {
```
jexp: we should also consider a bidirectional approach

jexp: TODO compare with built-in dijkstra

PR #64, opened 2017-04-10
Api2 first draft

mknblch:
The pr aims to upgrade the current api to the new approach. Basically we split the graph into several interfaces to improve granularity of local data. I also added 2 implementations for the AllRelationIterator-iface, an IdMapper based on hppc maps and the int-long-converter from #56.


core/src/main/java/org/neo4j/graphalgo/core/sources/SingleRunAllRelationIterator.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class SingleRunAllRelationIterator implements AllRelationIterator, AutoCloseable {
+
+    private final GraphDatabaseAPI api;
+    private final Transaction transaction;
+    private final IdMapping idMapping;
+
+    public SingleRunAllRelationIterator(GraphDatabaseAPI api, IdMapping idMapping) {
+        this.api = api;
+        transaction = api.beginTx();
```
jexp: has to be ensured that this is only used in a single thread, better handle the transaction in the forEachRel... method

PR #66, opened 2017-04-11
Api2 data sources

mknblch:
Adds a datasource for a single relationship type which implements iteration in one direction.


core/src/main/java/org/neo4j/graphalgo/core/utils/RelationshipContainer.java
```diff
@@ -0,0 +1,215 @@
...
+         * set the builder to aim for nodeId and initialize its underlying connection array
+         *
+         * @param nodeId the current (mapped) nodeId
+         * @param degree the degree for the node
+         * @return this instance for method chaining
+         */
+        public Builder aim(int nodeId, int degree) {
+            this.nodeId = nodeId;
+            offset = 0;
+            container.data[nodeId] = new int[degree];
```
jexp: could the degree also be inferred upon adding new targetNodeIds ? (at least as an option?

mknblch: I could add a logic which grows the array on demand.

jexp: Oh sorry, I thought this was just keeping track of the degree counter.
If we have the data upfront, it's fine to pre-size it.

mknblch: Ok - But you are right. We might need an autogrowing implementation in future for extracting paths (e.g. shortest path forest)

mknblch: Or perhaps we find an even better struct for this case. I made issue #69 out of it but im going to delay it until we need it.

PR #74, opened 2017-04-20
Procedures 1st draft

mknblch:
- the PR adds 2 procedures for streaming unions based on the Graph-Impl. 
- adaptions in GraphLoader to ease the use in procedures. 
- DSS got some methods for iterating and streaming
- simplified UnionFind


algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class UnionFindProc {
+
+    @Context
+    public GraphDatabaseAPI api;
+
+    @Context
+    public Log log;
+
+    @Procedure(value = "org.neo4j.graphalgo.unionFind")
```
jexp: simplify the name to algo.unionFind

algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
```diff
@@ -0,0 +1,64 @@
...
+    @Description("CALL org.neo4j.graphalgo.unionFind(label:String, relationship:String) " +
+            "YIELD nodeId, setId - yields a setId to each node id")
+    public Stream<DisjointSetStruct.Result> unionFind(
```
jexp: same for all other algorithms:

default values for label and rel-type of null (empty string), to run the algo across the whole graph

can be done later after this is merged

algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
```diff
@@ -0,0 +1,64 @@
...
+
+        return new GraphUnionFind(graph)
+                .compute()
+                .resultStream(graph);
+    }
+
+    @Procedure(value = "org.neo4j.graphalgo.unionFindThreshold")
+    @Description("CALL org.neo4j.graphalgo.unionFindThreshold(label:String, relationship:String, property:String, threshold:double) " +
+            "YIELD nodeId, setId - yields a setId to each node id")
+    public Stream<DisjointSetStruct.Result> unionFindThreshold(
```
jexp: merge it into one procedure and use default values for property (null) and threshold (0 or -1) to indicate that they are not set.

algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
jexp: perhaps flip label + rel-type ? put the one first that's more common

algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
```diff
@@ -0,0 +1,108 @@
...
+    public GraphDatabaseAPI api;
+
+    @Context
+    public Log log;
+
+    @Procedure(value = "algo.unionFind")
+    @Description("CALL algo.unionFind(label:String, relationship:String, {property:'propertyName', threshold:0.42, defaultValue:1.0}) " +
+            "YIELD nodeId, setId - yields a setId to each node id")
+    public Stream<DisjointSetStruct.Result> unionFind(
+            @Name("label") String label,
```
jexp: label and rel-type should still have a defaultValue=""
if you want to run it on the whole graph

algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
```diff
@@ -0,0 +1,108 @@
...
+            return new GraphUnionFind(graph)
+                    .compute()
+                    .resultStream(graph);
+        }
+    }
+
+    @Procedure(value = "algo.unionFindStats")
+    @Description("CALL algo.unionFindStats(label:String, relationship:String, {property:'propertyName', threshold:0.42, defaultValue:1.0}) " +
+            "YIELD setId, setSize - yields the size for each setId")
+    public Stream<Result> unionFindStats(
```
jexp: Would it make sense to return this as a statistics result with runtime, number of nodes / rels touched, number of clusters and the {setId:setSize} as a map? And also handle writing back here, i.e. if the config contains write:true it writes back the "clusterId" property (or whatever is configured in "clusterProperty" to the nodes.

jexp: If someone wants to return the results as a table then, they can still do:

```
UNWIND keys(result.clusters) as cluster
RETURN cluster, result.clusters[cluster] as size
```

jexp: @mknblch I can merge the current state and we adapt the result later on, or you can to the statistics result (similar to apoc's pagerankstats) first?

algo/src/main/java/org/neo4j/graphalgo/UnionFindProc.java
jexp: for config we can use also defaultValue=""

PR #75, opened 2017-04-20
Add data structure for undirected graph represented as trees

knutwalker:
This one will be useful for operation on spanning trees and other cluster related things.

@mknblch check out the tests for usage and see if you need any API changes.


core/src/main/java/org/neo4j/graphalgo/core/utils/container/UndirectedTree.java
```diff
@@ -0,0 +1,269 @@
...
+                            + capacity);
+        }
+    }
+
+    private Boolean verify(int node1, int node2) {
+        if (node1 == node2) {
+            // self relation, nothing to do
+            return null;
+        }
+        verify(node1);
```
jexp: isn't this expensive to do every time?

PR #91, opened 2017-04-24
Use proper casts and config changes for page rank proc

### LINKED ISSUE #88: calling page-rank-proc with iterations fails
CALL algo.pageRankStream('Person','KNOWS',{iterations:5}) yield node, score
> WITH * order by score DESC LIMIT 10
> RETURN id(node), node.firstName, node.lastName, score;
239 ms

WARNING: Failed to invoke procedure `algo.pageRankStream`: Caused by: java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Integer

### LINKED ISSUE #90: pageRankStats cannot write
```
neo4j-sh (?)$ CALL algo.pageRankStats('Person','KNOWS',{write:true});
7 ms

WARNING: Failed to invoke procedure `algo.pageRankStats`: Caused by: java.lang.UnsupportedOperationException

```

knutwalker:
fixes #88
fixes #90

Regarding #90 though: only pageRank saves, pageRankStats only returns stats as a shorthand for pageRank with write:false. Might just remove the stats one, if that's confusing.


algo/src/main/java/org/neo4j/graphalgo/PageRankProc.java
```diff
@@ -86,11 +87,18 @@
             @Name(value = "label", defaultValue = "") String label,
             @Name(value = "relationship", defaultValue = "") String relationship,
             @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
-        config.put(CONFIG_WRITE, Boolean.FALSE);
+        if ((boolean) config.getOrDefault(CONFIG_WRITE, DEFAULT_WRITE)) {
+            Map<String, Object> newConfig = new HashMap<>(config);
+            newConfig.put(CONFIG_WRITE, Boolean.FALSE);
+            config = newConfig;
+        }
         return pageRank(label, relationship, config);
```
jexp: for later, the internal methods that are called should just have spelled out parameters, so we just pass the right ones along.

PR #92, opened 2017-04-25
Fixuf

mknblch:
- Fix for DssExporter when creating unknown properties (e.g. for cluster id)
- added another testcase for uninitialized DSS


core/src/main/java/org/neo4j/graphalgo/core/utils/dss/DisjointSetStruct.java
```diff
@@ -291,7 +292,13 @@ public Cursor next() {
         public DSSExporter(GraphDatabaseAPI api, IdMapping idMapping, String targetProperty) {
             super(api);
             this.idMapping = idMapping;
-            propertyId = getOrCreatePropertyId(targetProperty);
+            writeInTransaction(write -> {
```
jexp: shouldn't that go into `getOrCreatePropertyId` ?

PR #104, opened 2017-04-27
Implementing Cypher Loading. First for Heavy Graph

jexp:


jexp:
Changed GraphSetup to have an interface, making the way to have different implementations.
First direct-loading, no parallelization.

core/src/main/java/org/neo4j/graphalgo/api/CypherGraphSetup.java
```diff
@@ -0,0 +1,131 @@
...
+
+import org.neo4j.kernel.internal.Version;
+
+import java.util.concurrent.ExecutorService;
+
+/**
+ * @author mh
+ * @since 27.04.17
+ */
+public class CypherGraphSetup implements GraphDefinition {
```
knutwalker: I'm not sure if we really need the two implementations.
If we add the two statement properties to the GraphLoader API, we can keep the one impl GraphSetup/Definition situation, have the CyperFactory extends GraphFactory as well and use the same API for all loaders, which would allow us to include it easily in tests like GraphNegativeTest.
Since loading only happens once, we can do some validation, maybe throw an IllegalState|ArgumentException, when defining both, Label/RelTypes and Cypher Statements.

WDYT?

core/src/main/java/org/neo4j/graphalgo/core/heavyweight/HeavyCypherGraphFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        int nodeCount = idMap.size();
+
+        final AdjacencyMatrix matrix = new AdjacencyMatrix(nodeCount);
+
+        final WeightMapping relWeigths = hasRelationshipWeights ?
+                new WeightMap(nodeCount * DEGREE, setup.relationDefaultWeight()) :
+                new NullWeightMap(setup.relationDefaultWeight());
+
+        api.execute(setup.relationshipStatement()).accept(row -> {
+            int source = idMap.mapOrGet(row.getNumber("source").longValue());
```
knutwalker: Could we assume that the node statement always loads all nodes we need and use `get` here, skipping the rel or emit a warning, when `get` returns -1?

core/src/main/java/org/neo4j/graphalgo/api/CypherGraphSetup.java
knutwalker: For example, tests like [GraphNegativeTest](https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/32ff4065dbb324c5965d44718a8e111f2e716086/tests/src/test/java/org/neo4j/graphalgo/core/GraphNegativeTest.java#L36-L43), [PageRankTest](https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/32ff4065dbb324c5965d44718a8e111f2e716086/tests/src/test/java/org/neo4j/graphalgo/impl/PageRankTest.java#L32-L39), or [ShortestPathDijkstraTest](https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/32ff4065dbb324c5965d44718a8e111f2e716086/tests/src/test/java/org/neo4j/graphalgo/impl/ShortestPathDijkstraTest.java#L29-L36) run via junit parameters on all graph impls and we could include cypher loading there as well.

core/src/main/java/org/neo4j/graphalgo/core/heavyweight/HeavyCypherGraphFactory.java
jexp: right we should use get here and skip it then

core/src/main/java/org/neo4j/graphalgo/core/heavyweight/AdjacencyMatrix.java
```diff
@@ -93,7 +93,7 @@ public void armIn(int targetNodeId, int degree) {
     public void addOutgoing(int sourceNodeId, int targetNodeId, long relationId) {
         final int degree = outOffsets[sourceNodeId];
         final int nextDegree = degree + 1;
-
+        if (outgoing[sourceNodeId].length <= degree) armOut(sourceNodeId, degree+Math.min(Math.max(degree,10)<<2,2<<16));
```
mknblch: please use brackets due to consistency

knutwalker:
LGTM

PR #106, opened 2017-04-27
Algorithm: Betweenness Centrality

mknblch:
The PR adds an efficient BetweennessCentrality algorithm as described in [this paper](http://www.algo.uni-konstanz.de/publications/b-fabc-01.pdf).


algo/src/main/java/org/neo4j/graphalgo/results/AbstractResult.java
```diff
@@ -1,42 +0,0 @@
-package org.neo4j.graphalgo.results;
```
mknblch: This was a test which didn't work well

core/src/main/java/org/neo4j/graphalgo/core/utils/container/Path.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.lucene.util.ArrayUtil;
+
+import java.util.Arrays;
+import java.util.function.IntPredicate;
+
+/**
+ * @author mknblch
+ */
+public class Path {
```
mknblch: PoC Impl. Lets find a better container for shortest path forests.

mknblch:
@jexp the PR can be merged now.

PR #112, opened 2017-05-03
Procedure: SCCTarjan

mknblch:
- add Impl + Proc + Test for SCCTarjan
- add AbstractResultBuilder to ease development of new result builders


algo/src/main/java/org/neo4j/graphalgo/SCCTarjanProc.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Map;
+import java.util.stream.Stream;
+
+/**
+ * @author mknblch
+ */
+public class SCCTarjanProc {
+
+    public static final String CONFIG_WRITE = "write";
```
jexp: should go into some central Config class at some point that we can use to turn the maps in proper config objects

algo/src/main/java/org/neo4j/graphalgo/SCCTarjanProc.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final String CONFIG_WRITE_PROPERTY = "clusterProperty";
+    public static final String CONFIG_CLUSTER = "cluster";
+
+    @Context
+    public GraphDatabaseAPI api;
+
+    @Context
+    public Log log;
+
+    @Procedure(value = "algo.scctarjan", mode = Mode.WRITE)
```
jexp: the algorithm procedure should just be called algo.scc or algo.stronglyConnectedComponents

Tarjan is an implementation detail.

jexp: Please fix in a second PR

PR #113, opened 2017-05-03
Graphbuilder Testhelper

mknblch:
Often i need to define graphs for unit tests. this commit adds more GraphBuilders to create testgraphs with predefined properties.

- LineBuilder: create line of connected nodes
- GridBuilder: creates a grid of nodes
- CompleteGraphBuilder: create graph where each node is connected to any other node

@jexp could we benefit from having procedures for the graphBuilders?


tests/src/main/java/org/neo4j/graphalgo/core/graphbuilder/CompleteGraphBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author mknblch
+ */
+public class CompleteGraphBuilder extends GraphBuilder<CompleteGraphBuilder> {
+
+    protected CompleteGraphBuilder(GraphDatabaseAPI api, Label label, RelationshipType relationship) {
+        super(api, label, relationship);
+    }
+
+    public CompleteGraphBuilder createCompleteGraph(int nodeCount) {
+        ArrayList<Node> nodes = new ArrayList<>();
```
jexp: List<Node>

tests/src/main/java/org/neo4j/graphalgo/core/graphbuilder/GraphBuilder.java
```diff
@@ -22,6 +22,7 @@
     private final ME self;
 
     protected final HashSet<Node> nodes;
+    protected final HashSet<Relationship> relationships;
```
jexp: Set<Relationship>

Issue #68, opened 2017-04-11
Change Consumer into Functions to allow premature termination


mknblch:
Since we might decide to get rid of the Iterator-methods (in #29) we should add the possibility to terminate the iteration within a forEach(..) method before all values have been emitted. This could be implemented by changing the Consumer into Functions which return a Boolean that either stops or continues the current iteration.

jexp:
Yes BooleanFunction (i.e. Prediate).

mknblch:
`org.neo4j.graphalgo.api.RelationshipConsumer` has already been changed in 3.1. `org.neo4j.graphalgo.api.NodeIterator` still needs to be done.

Issue #29, opened 2017-03-24
Decide for one approach of Iterator vs Consumer vs Lookup in Graph API


knutwalker:
The Graph currently provides [multiple ways](https://en.wikipedia.org/wiki/There%27s_more_than_one_way_to_do_it) to access the graph data. We should decide for one with the consumer based API being the favorite.

https://github.com/neo4j-contrib/neo4j-graph-algorithms/pull/4#discussion_r106820797

PR #116, opened 2017-05-03
Changed NodeIterator to use IntPredicate instead of IntConsumer

mknblch:
This PR completes #68. I know this one was deferred but it would allow us to do #29 which in return would reduce the code-size/test-duration significantly. Between all the current pullrequests I would suggest to merge this one first. I can easily adapt the others afterwards.


jexp:
I merge it for now, but we should def. change this later for a proper NodeConsumer interface, not just a generic one that doesn't communicate what that "int" is.

PR #122, opened 2017-05-05
Parallel mst import

mknblch:
Changes DataImport to perform loading of weights and relationships in parallel


mknblch:
merged with 3.1 which reduced size of the diff

algo/src/main/java/org/neo4j/graphalgo/MSTPrimProc.java
```diff
@@ -43,36 +46,48 @@
...
-        ProgressTimer timer = ProgressTimer.start(builder::withLoadDuration);
-        BufferedWeightMap weightMap = BufferedWeightMap.importer(api)
+        CompletableFuture<BufferedWeightMap> weightMap = BufferedWeightMap.importer(api)
                 .withIdMapping(idMapper)
                 .withAnyDirection(true)
                 .withWeightsFromProperty(propertyName, 1.0)
                 .withOptionalLabel(configuration.getNodeLabelOrQuery())
                 .withOptionalRelationshipType(configuration.getRelationshipOrQuery())
-                .build();
+                .buildDelayed(Pools.createDefaultPool());
```
jexp: shouldn't we use a single pool to avoid contention?

algo/src/main/java/org/neo4j/graphalgo/MSTPrimProc.java
```diff
@@ -43,36 +46,48 @@
...
+                .withWeightsFromProperty(propertyName, 1.0)
+                .withOptionalLabel(configuration.getNodeLabelOrQuery())
+                .withOptionalRelationshipType(configuration.getRelationshipOrQuery())
+                .build();
 
         MSTPrimResult.Builder builder = MSTPrimResult.builder();
 
-        ProgressTimer timer = ProgressTimer.start(builder::withLoadDuration);
-        BufferedWeightMap weightMap = BufferedWeightMap.importer(api)
+        CompletableFuture<BufferedWeightMap> weightMap = BufferedWeightMap.importer(api)
```
jexp: I think we should move the code out of the procedure into the algorithm or into a dedicated (reusable?) parallel loader?

jexp:
@mknblch did you test how it performs across CPUs? vs. the previous version?

algo/src/main/java/org/neo4j/graphalgo/MSTPrimProc.java
mknblch: Yea sure. The import logic of dedicated data sources is far from beeing complete. But it was obvious that we can do the imports of several data sources in parallel. Up to now the logic is implemented in the abstract Importer class. But since the MST procedure is the only one which uses the dedicated api2 sources we should defer this to later.

PR #124, opened 2017-05-07
Proc followup

mknblch:
The PR unifies resultBuilders for PageRank, BetweennessCentrality, UnionFind, LabelPropagation. Also made ProcessConfiguration writable bc the map given from procedure throws an exception upon write.


algo/src/main/java/org/neo4j/graphalgo/PageRankProc.java
```diff
@@ -18,21 +19,16 @@
...
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 public final class PageRankProc {
 
     public static final String CONFIG_DAMPING = "dampingFactor";
-    public static final String CONFIG_ITERATIONS = "iterations";
-    public static final String CONFIG_WRITE = "write";
-    public static final String CONFIG_SCORE_PROPERTY = "scoreProperty";
     public static final String BATCH_SIZE_PROPERTY = "batchSize";
```
jexp: batchSize should also go into the config

algo/src/main/java/org/neo4j/graphalgo/LabelPropagationProc.java
```diff
@@ -18,18 +19,14 @@
...
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
 import static algo.util.Util.parseDirection;
 
 public final class LabelPropagationProc {
 
-    public static final String CONFIG_ITERATIONS = "iterations";
-    public static final String CONFIG_WRITE = "write";
     public static final String CONFIG_WEIGHT_KEY = "weightProperty";
```
jexp: weight and partitionProperty should imho also go into the general config

algo/src/main/java/org/neo4j/graphalgo/LabelPropagationProc.java
```diff
@@ -18,18 +19,14 @@
...
     public static final String CONFIG_PARTITION_KEY = "partitionProperty";
-
     public static final Integer DEFAULT_ITERATIONS = 1;
     public static final Boolean DEFAULT_WRITE = Boolean.TRUE;
     public static final String DEFAULT_WEIGHT_KEY = "weight";
```
jexp: default weight key should also go into the default config

algo/src/main/java/org/neo4j/graphalgo/StronglyConnectedComponentsProc.java
```diff
@@ -21,7 +21,6 @@
  */
 public class StronglyConnectedComponentsProc {
 
-    public static final String CONFIG_WRITE = "write";
     public static final String CONFIG_WRITE_PROPERTY = "clusterProperty";
```
jexp: is partition the same as cluster ? should we merge the two namings ?

jexp:
Just curious @mknblch . When do you need to write to the map?

PR #134, opened 2017-05-12
parallel Delta-stepping Non-negative Single Source Shortest Paths

mknblch:
This is an alternative algorithm for shortest paths. According to different papers it's parallel performance is better then parallel-dfs implementations so I gave it a try.

The deltaStepping algorithm can be tweaked with the delta parameter ([more](https://ae.cs.uni-frankfurt.de/pdf/diss_uli.pdf)).

Based on this implementation we can build parallel centrality measurement algorithms easily. 

TODO:
- edge case tests
- benchmark


jexp:
@mknblch is this ready to be merged?

PR #135, opened 2017-05-12
delta stepping nsssp & procedure & test

mknblch:
This is an alternative algorithm for non-negative single source shortest paths. According to different papers it's parallel performance is better then common parallel-dfs implementations so I gave it a try. [More](https://ae.cs.uni-frankfurt.de/pdf/diss_uli.pdf) or [here](https://arxiv.org/pdf/1604.02113v1.pdf).

There is still potential for optimization.  

- Simplified PR

**TODO**
- benchmark
- edge case test
- test on bigger graph


jexp:
@mknblch is this ready to be merged?

mknblch:
@jexp after reviewing the implementation im unsure about it's behaviour on bigger graphs in parallel. We do concurrent access to Buckets and the distance[] array. According to the papers the Buckets should be safe but we might need to switch to an atomic array for the distances. I was not able to produce any failures on the smal graph but it may happen on bigger ones with cycles etc..

Anyways.. The sequential usage is fine and i already have a branch with ClosenessCentrality based on this SSSP algo. I deactivated the parallel evaluation in the procedure so we can benchmark its sequental performance against dijkstra.

mknblch:
But i have no idea why `shouldGetPageRankWithCypherExpectedResultWithoutNodesParallel(algo.algo.PageRankTest)` fails. I did not change anything there.

PR #146, opened 2017-05-16
Ldbc benchmark for LabelPropagation

### LINKED ISSUE #141: Finish benchmark PR for LabelPropagation

knutwalker:
Closes #141 

Also, add execution date to benchmark result file names.


benchmark/src/main/java/org/neo4j/graphalgo/bench/LabelPropagationBenchmarkLdbc.java
```diff
@@ -0,0 +1,144 @@
...
+    public void failTx() {
+        tx.failure();
+        tx.close();
+    }
+
+    @Benchmark
+    public Object _01_algo() {
+        return runPrintQuery(
+                db,
+                "CALL algo.labelPropagation(null, null, 'OUTGOING',{iterations:" + iterations + ",batchSize:" + batchSize + "}) "
```
jexp: for the future we should use parameters here, just to have a nicer call

benchmark/src/main/java/org/neo4j/graphalgo/bench/LdbcDownloader.java
```diff
@@ -0,0 +1,192 @@
+package org.neo4j.graphalgo.bench;
```
jexp: a one line shell script would have done the same :)
```
curl -s -L -o - 'http://example-data.neo4j.org.s3.amazonaws.com/files/ldbc_sf001_p006_neo4j31.tgz' | $(export DIR="$TMPDIR$$/L01" && mkdir -p $DIR && tar -xf - -C $DIR && echo $DIR)
```

PR #150, opened 2017-05-18
rm parallel import due to nondeterministic tests

mknblch:
Sometimes i notice an IndexArrayOutOfBoundsException when running MSTPrimTest in Maven. We should review the implementation and test the parallel import. Up to now i deactived the parallel import.


jexp:
@mknblch shouldn't it still be deterministic in the results?

What happens to users that use the parallel import, will they see vastly different results?

PR #151, opened 2017-05-18
Delta stepping speedup

mknblch:
This PR reduces some overhead from scaling the values in DeltaSteppingAlgo. It also adds a benchmark. Currently the benchmark produces strange values. We should check the parallel impl.


algo/src/main/java/org/neo4j/graphalgo/ShortestPathDeltaSteppingProc.java
```diff
@@ -70,7 +72,9 @@
                 .load(HeavyGraphFactory.class);
 
         return new ShortestPathDeltaStepping(graph, delta)
-                .withExecutorService(Pools.DEFAULT)
+                .withExecutorService(Executors.newFixedThreadPool(
+                        configuration.getInt("concurrency", 4)
```
jexp: the concurrency default should come from the config and there it should be inferred from the available processors or global configuration.

jexp:
@mknblch can you do the concurrency config default I mentioned in a separate PR ?

PR #154, opened 2017-05-20
PageRank and LabelPropagation syntax clean

tomasonjo:
changes made
Label propagation:
  Input:
      -partitionProperty -> writeProperty
  Output:
     -loadMillis -> loadDuration
     -writeMillis -> writeDuration
     -computeMillis -> evalDuration
     -partitionProperty -> writeProperty

PageRank:
  Input:
    - add writeProperty in procedure description
    - Should we change scoreProperty to weightProperty ?
 
 Output:
    -property -> writeProperty
    -loadMillis -> loadDuration
    -computeMillis -> evalDuration
    -writeMillis -> writeDuration


jexp:
I still think we have to communicate the time-unit somewhere for the user.

And personally I prefer compute over eval, as compute expresses more that there is a full processing algorithm running.

I think changing things consistently to writeProperty makes sense.
How much do you think we have to say `writePropertyName`

tomasonjo:
I too prefer compute over eval as it makes more sense.

Do not think we gain much by changing `writeProperty`  to `writePropertyName`. It does not add any more clarification than `writeProperty`. Best to keep it short in my opinion for better user experience. 

To me  `loadDuration` makes more sense than `loadMillis`. Would not go for `loadDurationMillis` even though it looks cool, because I do not like typing long variables as a user. But I am open to suggestion  to change `Duration` to `Millis` where we always communicate time-unit.
If we keep `Duration`, user can check the syntax part of docs. Should not be a big problem as it is always milliseconds.

tomasonjo:
Ok so if we want to communicate the time-unit, we should standardize the following parameters:
weightProperty
writeProperty
loadMillis
computeMillis
writeMillis

tomasonjo:
I reverted back to Millis format.
Final Changes:
  Label propagation:
    Input:
        -partitionProperty -> writeProperty
    Output:
         -property -> writeProperty
  PageRank:
    Input:
        - add writeProperty in procedure description
    Output:
        -property -> writeProperty
Suggestions:
  --If pageranks scoreProperty is actually a weightProperty we should rename it. I did not want to fiddle around with variables too much, because i am Java Noob :)
  --Change all durations output to Millis in other algos

knutwalker:
The partition key for label propagation is not just used as an output property but also to read existing values during computation, `writeProperty` would be confusing there. The best solution would probably to add the read property back as partitionKey/Property, defaulting to the same value and use this one during load and `writeProperty` during write.

tomasonjo:
Ok I reverted also the Label Prop.
Only changes left are:
PageRank:
 -add writeProperty as input in description
 -change property output to writeProperty

I think it is best if I raise issues for pageRank scoreProperty and Durations property, so I do not mess with your algos too much :)

jexp:
Thanks a lot

PR #157, opened 2017-05-23
docs_examples

tomasonjo:
Added a basic syntax for all:

Two things come to mind:
 - PageRank has scoreProperty and writeProperty that both default to "score", which is confusing
 - UnionFind has property parameter that should be a weightProperty i guess


jexp:
@tomasonjo is this ready to merge?

tomasonjo:
Yes it is fine to merge so that we show all the syntax for now. Need to correct some syntax in a different PR for more clarity next I think.

doc/connected-components.adoc
```diff
@@ -12,10 +18,79 @@ _Connected Components_ or _UnionFind_ basically finds sets of connected nodes wh
...
+| relationship | string | null | yes | relationship-type to load from the graph, if null load all relationships
+
+|===
+
+.Results
+[opts="headers"]
+|===
+| name | type | description
+| nodeId | long | node id
+| setId | int | cluster id
```
jexp: there is no difference between long and `int` we should use `integer` and `float` (i.e. the cypher types) consistently throughout the docs.

doc/label-propagation.adoc
```diff
@@ -1,22 +1,66 @@
...
+[opts="header",cols="1,1,6"]
+|===
+| name | type | description
+| nodes | int | number of nodes considered
+| iterations | int | number of iterations
+| loadMillis | int | milliseconds for loading data
+| computeMillis | int | milliseconds for running the algorithm
+| writeMillis | int | milliseconds for writing result data back
+| weightProperty | string | property name that contains weight
+| partitionProperty | string | property name written back to
```
jexp: what is written back

doc/label-propagation.adoc
```diff
@@ -1,22 +1,66 @@
...
+
+|===
+
+.Results
+[opts="header",cols="1,1,6"]
+|===
+| name | type | description
+| nodes | int | number of nodes considered
+| iterations | int | number of iterations
+| loadMillis | int | milliseconds for loading data
```
jexp: include from common results table

doc/minimum-weight-spanning-tree.adoc
```diff
@@ -1,24 +1,76 @@
...
+There are quite a few use cases for minimum spanning trees. One example would be a telecommunications company which is trying to lay out cables in new neighborhood. If it is constrained to bury the cable only along certain paths (e.g. along roads), then there would be a graph representing which points are connected by those paths. Some of those paths might be more expensive, because they are longer, or require the cable to be buried deeper; these paths would be represented by edges with larger weights. Currency is an acceptable unit for edge weight – there is no requirement for edge lengths to obey normal rules of geometry such as the triangle inequality. A spanning tree for that graph would be a subset of those paths that has no cycles but still connects to every house; there might be several spanning trees possible. A minimum spanning tree would be one with the lowest total cost, thus would represent the least expensive path for laying the cable.
+
 == Constraints / when not to use it
 
 == Algorithm explanation on simple sample graph
 
 [source,cypher]
 ----
+MATCH(n:Node{start:true}) 
+WITH n
```
jexp: we don't need the `WITH n`

doc/minimum-weight-spanning-tree.adoc
```diff
@@ -1,24 +1,76 @@
...
+
 == Constraints / when not to use it
 
 == Algorithm explanation on simple sample graph
 
 [source,cypher]
 ----
+MATCH(n:Node{start:true}) 
+WITH n 
+CALL algo.mst(n, 'cost', {write:true, stats:true,writeProperty:"mst"})
```
jexp: do we need stats:true in the stats procedure?

doc/pagerank.adoc
```diff
@@ -31,26 +34,33 @@ CALL algo.pageRank(label:String, relationship:String, {iterations:5, dampingFact
 | name | type | default | optional | description
 | label  | string | null | yes | label to load from the graph, if null load all nodes
 | relationship | string | null | yes | relationship-type to load from the graph, if null load all nodes
-| iterations | int | 5 | yes | how many iterations of page-rank to run
+| iterations | int | 20 | yes | how many iterations of page-rank to run
 | dampingFactor | float | 0.85 | yes | damping factor of the page-rank caculation
 | write | boolean | true | yes | if result should be written back as node property
-| scoreProperty | string | 'score' | yes | property name written back to
+| writeProperty | string | 'score' | yes | property name written back to
+| scoreProperty | string | 'score' | yes | property name that contains weight
```
jexp: should both scoreProperty and writeProperty have the same value? shouldn't writeProperty be sth like "rank" or "pageRank" ?

jexp: or "scoreProperty" should be "weigthProperty" or "propertyName" ?

doc/single-shortest-path.adoc
```diff
@@ -14,6 +14,34 @@ _All Pairs Shortest Path_ on the other hand calculates a shortest path forest co
...
+----
+MATCH (start:Node{type:'start'}), (end:Node{type:'end'})
+CALL algo.shortestPath(start, end, 'cost') 
+YIELD loadDuration, evalDuration, nodeCount, totalCost
+
+----
+
+[source,cypher]
+----
+MATCH(n:Node {name:'s'}) WITH n CALL algo.deltaStepping.stream(n, 'cost', 3.0)
```
jexp: @mknblch should we expose implementation details in the procedure name? Can we find the better "common" algorithm name that's covered here?

doc/strongly-connected-components.adoc
```diff
@@ -12,10 +14,46 @@ _SCC_ is a class algorithms for finding groups of nodes where each node is direc
...
 ----
 
 == Example Usage
 
+== Syntax
+
+.Running algorithm and writing back results
+[source,cypher]
+----
+CALL algo.scc(label:String, relationship:String, {write:true,clusterProperty:'cluster'})
```
jexp: @mknblch missing writeProperty ? both as config and result?

tomasonjo: clusterProperty is the "writeProperty" in this scenario

doc/pagerank.adoc
tomasonjo: Yes I think that it should be different defaults... i went with what I read from the code... I think that "scoreProperty" is "weightProperty" not sure... we should avoid using propertyName as it is confusing imo

doc/connected-components.adoc
```diff
@@ -12,10 +18,79 @@ _Connected Components_ or _UnionFind_ basically finds sets of connected nodes wh
...
+
+----
+
+.Parameters
+[opts="header",cols="1,1,1,1,4"]
+|===
+| name | type | default | optional | description
+| label  | string | null | yes | label to load from the graph, if null load all nodes
+| relationship | string | null | yes | relationship-type to load from the graph, if null load all nodes
+| property | string | null | yes | property name that contains weight
```
tomasonjo: is there any difference in weight vs cost (I think not), both are properties of rels ? should just pick one and be consistent... avoid property as it is confusing

doc/label-propagation.adoc
tomasonjo: The partition of the graph in which the node resides

doc/connected-components.adoc
```diff
@@ -12,10 +18,79 @@ _Connected Components_ or _UnionFind_ basically finds sets of connected nodes wh
...
 ----
 
 == Example Usage
 
+== Syntax
+
+.Running algorithm and writing back results
+[source,cypher]
+----
+CALL algo.unionFind(label:String, relationship:String, {property:'propertyName', threshold:0.42,
```
tomasonjo: should change also the `property` to `weightProperty`, just spoke with martin and he said that whenever the default value for `property` is cost or the name of the param is `costProperty` it means the same as `weightProperty`. So I would stick with `weightProperty`

jexp: Ok, makes sense, as long as we are consistent, and it makes sense for the user.

doc/connected-components.adoc
jexp: Yes. Can you work with @mknblch  on this?

doc/label-propagation.adoc
jexp: Put it in :)

doc/minimum-weight-spanning-tree.adoc
jexp: I mean generally.

doc/pagerank.adoc
jexp: Please ask @mknblch

doc/strongly-connected-components.adoc
jexp: Then let's change it.

jexp:
@tomasonjo should I merge this in and then you work on my comments in another round?

tomasonjo:
No please do not merge. Will push commits today afternoon.

tomasonjo:
added some examples and updated the syntax according to [#160](https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/160)

PR #163, opened 2017-05-31
[WIP] parallel connected components & thresholding CC

mknblch:
WIP!

I added a parallel unionfind implementation. The benchmark shows a speedup to sequential UF at about 1m nodes & 1m relations on my machine. I adapted the Procedure so that it uses parallelUF if a batchsize is given, sequential UF otherwise (we might change this!). Unfortunately I can't create TestGraphs with more than 1m nodes in reasonably time. We should benchmark it on bigger graphs with sufficient relationships. 

I can't exactly tell where i got the idea for it (was a paper of an US univerity). It works by partitioning the nodeset into batches wich are processed in parallel. Each process computes the disjointset-structure for it's nodes. The resulting structs are then merged pairwise until one struct is left which is also done in parallel. 

Since PUF creates a DSS for each thread in nodeCount-size i also adapted DSS such that its arrays can be initialized with a single value (Arrays.fill). 

**TODO**
- better way for controlling paralellism 
- Edge case Tests for DSS.merge(..)
- Copy Pools impl. into algo-package. 
- Add default forkJoinPool 
OR
- replace ForkJoin logik

**Improvement**
The DSS-computation is currently made with ParallelUtil which waits until all tasks are completed. The merge process uses a ForkJoin Logik (RecursiveTask). The profiling shows a big drop in CPU usage if the Util is waiting for all threads to complete. We could also do the DSS computation in the ForkJoin Task itself - i expect a performance gain.

Up to now I split the nodeset based on a batchSize this splits the computation into N processes while N may not be the the optimal thread count.


jexp:
Cool sounds like a great, reasonable idea.
Let me know when you're done.

I can have a look at the test data creation which should actually not take that long.
I think should switch from impermanent-dbs for those tests to real ones.

PR #165, opened 2017-06-02
Configurable graph implementation

knutwalker:
PageRank is implemented, remaining algos will follow.

Should we also document the parameter or leave it undocumented?


knutwalker:
Most Algos support the `graph` parameter. The two that don't are Label Prop b/c is requires the HeavyGraph and MSTPrim, b/c it doesn't uses the new source based API.

Most tests run with Heavy and Light impls. I removed the View/Kernel impl b/c it's broken and produces many test errors (#166). I didn't implement the cypher impl b/c it must be called differently.

The question about documentation remains: should we document this parameter or keep it hidden?

jexp:
@knutwalker we keep it hidden for now.

PR #167, opened 2017-06-02
parallel Connected Components

mknblch:
i added 3 different impl. of parallel union find

**1. algo.unionFind** 
uses sequential impl. 

**2. algo.unionFind.exp1**
uses the queeue bases impl. ParallelUnionFindQueue

**3. algo.unionFind.exp2**
uses ParallelUtil for calculating the DSS & ForkJoin for merging (ParallelUnionFindFJMerge)

**4. algo.unionFind.exp3**
uses only fork join logic (ParallelUnionFindForkJoin)

Note: the parallel impl. is only active if a batchSize is given in config


jexp:
merged manually with fixes

PR #170, opened 2017-06-05
PageRank sample graph

tomasonjo:
I added the sample graph for pageRank with some history/explanation. Removed scoreProperty as it is not a `weightProperty`, but nothing as it was renamed to `writeProperty` and forgotten in description.


jexp:
LGTM. But isn't weightProperty properly implemented as weighted page-rank?

tomasonjo:
I have tested it on movies dataset and it shows no difference if I use weightProperty or not. I think this is the part of code that shows that there isnt a weighted page-rank yet.

 ```
private Graph load(
            String label,
            String relationship,
            Class<? extends GraphFactory> graphFactory,
            PageRankScore.Stats.Builder statsBuilder) {

        GraphLoader graphLoader = new GraphLoader(api)
                .withOptionalLabel(label)
                .withOptionalRelationshipType(relationship)
                .withoutRelationshipWeights()
                .withExecutorService(Pools.DEFAULT);

        try (ProgressTimer timer = statsBuilder.timeLoad()) {
            Graph graph = graphLoader.load(graphFactory);
            statsBuilder.withNodes(graph.nodeCount());
            return graph;
        }
    }
```

PR #172, opened 2017-06-07
Parallel asp

mknblch:
Implements a parallel AllShortestPath algorithm based on Dijkstra. Currently only the stream-variant is implemented.


algo/src/main/java/org/neo4j/graphalgo/AllShortestPathsProc.java
```diff
@@ -0,0 +1,60 @@
...
+        final Graph graph = new GraphLoader(api)
+                .withOptionalLabel(configuration.getNodeLabelOrQuery())
+                .withOptionalRelationshipType(configuration.getRelationshipOrQuery())
+                .withOptionalRelationshipWeightsFromProperty(
+                        propertyName,
+                        configuration.getPropertyDefaultValue(1.0))
+                .withExecutorService(Pools.DEFAULT)
+                .load(configuration.getGraphImpl());
+
+        return new AllShortestPaths(graph, Pools.DEFAULT, configuration.get("concurrency", 1))
```
jexp: default concurrency 1 ??

algo/src/main/java/org/neo4j/graphalgo/impl/AllShortestPaths.java
```diff
@@ -0,0 +1,142 @@
...
+        counter.set(0);
+        running = true;
+
+        for (int i = 0; i < concurrency; i++) {
+            executorService.submit(new ShortestPathTask());
+        }
+
+        long end = (long) nodeCount * nodeCount;
+
+        return LongStream.range(0, end)
```
jexp: close is unfortunately not called on these streams.

algo/src/main/java/org/neo4j/graphalgo/impl/AllShortestPaths.java
```diff
@@ -0,0 +1,142 @@
...
+                // scan relationships
+                graph.forEachRelationship(
+                        node,
+                        Direction.OUTGOING,
+                        (source, target, relId, weight) -> {
+                            // relax
+                            final double targetDistance = weight + sourceDistance;
+                            if (targetDistance < distance[target]) {
+                                distance[target] = targetDistance;
+                                queue.add(target, targetDistance);
```
jexp: this will return false if it was not added, and not block,
consider using offer() instead

knutwalker: `queue` is the thread local prio queue, not the shared blocking queue

algo/src/main/java/org/neo4j/graphalgo/impl/AllShortestPaths.java
knutwalker: The stream is bounded at any rate, so close is more like an early terminate, not a required thing

jexp: Ya this is something I had in other procedures like pulling data form an endpoint. That early termination (i.e. stop pulling data) doesn't do anything. Something I have to investigate in detail.

algo/src/main/java/org/neo4j/graphalgo/impl/AllShortestPaths.java
jexp: Ups sorry

PR #174, opened 2017-06-08
Adapt PR Formula

mknblch:
I adapted the formula in our PR implementation from

`(1 - d) / N + d * sum( PRj / DEGj) `

to

`(1 - d) + d * sum ( PRj / DEGj)`

We should get bigger values now. Lets compare them to APOC-impl


jexp:
@tomasonjo we might need to adapt the docs to this.

PR #177, opened 2017-06-09
[WIP] Parallel PageRank

knutwalker:
WIP: It's still a separate class, I want to merge this with existing PageRank first and remove the single threaded one, if the parallel implementation proves to be good enough. There are also explicit unit tests missing. The impl is currently tested in the proc test and passes all tests.

Please check the [JavaDoc](https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/2dff47d0381d77a26cee0f421e08dc4325838715/algo/src/main/java/org/neo4j/graphalgo/impl/ParallelPageRank.java#L29) for the implementation details.

This also pulls in the LMAX disruptor (which we wanted to use at some point anyways). If this is not desired, we could refactor this using plain old threads+queues.


jexp:
What was the reason for pulling in disruptor here but not in the other cases ?

Also I think disruptor is not best utilized if used in an ad hoc fashion. As you don't get the benefit of continuous ring buffers allocation. 

In my past work the ring buffer had to be quite large 

Von meinem iPhone gesendet

knutwalker:
@jexp you're correct, ring buffer is a bottle neck here. i have an update that is using an atomic array instead, performing better. I've also seen that this implementation performs better in the single thread case compared to our existing one, so we can profit at any rate.
I'll close the PR for now and update it later with the array changes, once I'm back from buzzwords (or find time there).

PR #178, opened 2017-06-15
Parallel pagerank

knutwalker:
Please check the [JavaDoc](https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/1d2a3dfdd17a3341f16831fb7781207107468879/algo/src/main/java/org/neo4j/graphalgo/impl/PageRank.java#L20) for the implementation details.

For the change in the Pool RejectionPolicy, this was the reason that the parallel implementation couldn't outperform the sequential one. On the small LDBC graph with default batch size, there are more partitions than threads and the policy was invoked often. With the 100ms sleep, this meant that the main thread running the algorithm was wasting time sleeping and the compute tasks finished long before that. Half of the time of the algorithm wasn't spent calculating anything, just waiting for the main thread to finish sleeping.
With the change to parking for 100 nano instead, the main thread spends its time in park and can run the synchronization and submit the computation tasks to the pool much more often. Using the CallerRunsPolicy would be even be more efficient, as the main thread would do computations instead of parking, but this only works out if the computation for the main thread are consistently smaller than the submitted ones. Here are two graph showing the change:

100ms sleep<br>
threads before with 100ms sleep<br>
100ns park<br>
threads after with 100ns park<br>


knutwalker:
Here are some numbers to compare. parallel implementation still doesn't really fly for me, but this can be investigated under another issue. These are single shot measurements on the LDBC graph.

|  impl |  iter |    par |     Score |
|-------|------:|-------:|----------:|
| Light |     5 |   true |  10.058 s |
| Light |     5 |  false |   9.372 s |
| Light |     5 |    old |  11.848 s |
| Light |    20 |   true |  12.234 s |
| Light |    20 |  false |  10.913 s |
| Light |    20 |    old |  23.089 s |
| Heavy |     5 |   true |  13.767 s |
| Heavy |     5 |  false |  13.197 s |
| Heavy |     5 |    old |  13.346 s |
| Heavy |    20 |   true |  16.539 s |
| Heavy |    20 |  false |  14.556 s |
| Heavy |    20 |    old |  15.019 s |

PR #179, opened 2017-06-19
(WIP) MultistepSCC

mknblch:
Still WIP please do not merge yet.

The PR contains the multistep-scc-algorithm implementation. The primary part (the coloring algorithm) is already parallized. The BFS, pivoting and trimming still work sequential. Im currently studying  the paper regarding MS-BFS with primitives because multistep could benefit from it. 

the prodecure is currently defined as "algo.scc.multistep" / "algo.scc.multistep.stream". This is the second SCC algo which produces the same result types as tarjanScc but takes a different set of configuration options. 

If you have suggestions regarding the procedure-name please let me know.


jexp:
Need to rebase after PR merge.
How's it going?

PR #180, opened 2017-06-19
Simplify and optimize PageRank for parallel use-case

knutwalker:
Use multiple partitioned arrays instead of a single shared Atomic array to avoid multiple threads contenting on the same slot. Since CAS is _only_ lock-free and not wait-free, only a single thread can make progress under contention. Best-case scenario is a completely balanced graph, but super nodes or important notes, as per this algorithm, are nodes that are calculated under high contention. In the end, such nodes are effectively calculated single threaded and are also responsible for the majority of calculations that have to be done. Enter Amdahl's Law and this explains, why the parallel implementation did see very little progress.

The new approach is explained in the [JavaDoc](https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/78ddd03a01b48a67783bc84e1bc2f533a6ada92f/algo/src/main/java/org/neo4j/graphalgo/impl/PageRank.java#L33-L60). The gist is, every partition (=thread) writes in it's own local block, a single thread reorders those blocks and for the next iteration, scores are calculated from the blocks of every other partition.

The effect is an increase in single-threaded run-time but an actual 2x improvement in the multi-threaded case.

<details>
<summary>Benchmark Results</summary>

*NOTE:* Just the calculation time, no loading. Loading adds too much uncertainty.

| graph | iterations | parallel | before (s) | after (s) |
|-------|-----------:|---------:|-----------:|----------:|
| LIGHT |          5 |     true |      0.888 |     0.449 |
| LIGHT |          5 |    false |      0.456 |     0.618 |
| LIGHT |         20 |     true |      3.502 |     1.470 |
| LIGHT |         20 |    false |      1.798 |     2.707 |
| HEAVY |          5 |     true |      0.780 |     0.375 |
| HEAVY |          5 |    false |      0.482 |     0.571 |
| HEAVY |         20 |     true |      3.143 |     1.453 |
| HEAVY |         20 |    false |      1.802 |     2.573 |
|  VIEW |          5 |     true |     26.806 |    14.893 |
|  VIEW |          5 |    false |     26.449 |    15.075 |
|  VIEW |         20 |     true |     90.268 |    54.558 |
|  VIEW |         20 |    false |     96.573 |    53.594 |

</details>


knutwalker:
@jexp yeah, on my laptop w/ 4 cores

knutwalker:
@jexp like to see how it does on your slightly bigger box :)

PR #182, opened 2017-06-23
extracted FWBW SCC algo from multistep into its own procedure

mknblch:


algo/src/main/java/org/neo4j/graphalgo/StronglyConnectedComponentsProc.java
```diff
@@ -140,4 +142,29 @@
         return multistep.resultStream();
     }
 
+
+    @Procedure(value = "algo.scc.fwbw.stream")
```
jexp: can we find a better name?

PR #183, opened 2017-06-26
parallel write for multistep + betweennessCentrality + unionFind

mknblch:
We now use ParallelExporter for all node-property based exports.

I also changed multistep to use a simple int-array as set-container instead of MapOfLists


jexp:
We'll need to rebase this PR after @knutwalker's has been merged.

PR #205, opened 2017-07-04
Add MS-BFS algorithm

knutwalker:
Implementation based on http://www.vldb.org/pvldb/vol8/p449-then.pdf, which is the backing paper of https://dl.dropboxusercontent.com/u/14493611/MS-BFS.pdf.

This implementation lends itself to high memory consumption, from the paper (Table 2):

<details>
<summary>memory consumption for given node and source counts</summary>

 Node Count | Source Count | Memory     |
-----------:|-------------:|-----------:|
  1,000,000 |           64 |    22.9 MB |
  1,000,000 |          512 |   183.1 MB |
  1,000,000 |        1,024 |   366.2 MB |
  1,000,000 |        4,096 |     1.4 GB |
  1,000,000 |        8,192 |     2.9 GB |
  1,000,000 |       32,768 |    11.4 GB |
 50,000,000 |        4,096 |    71.5 GB |
 50,000,000 |       32,768 |   572.2 GB |
 10,000,000 |   10,000,000 | 34924.6 GB |

</details>
<br/>

I'd like to investigate another paper (https://openproceedings.org/2017/conf/edbt/paper-107.pdf), that **might** promise a better solution.

This is a base for other algorithm, there is no procedure hook yet. A theoretical implementation for Closeness Centrality is described, which includes implementing a shortest path from all nodes to all nodes.


knutwalker:
One improvement could be that instead of creating all thread algorithm copies with their data structures upfront, we do this lazily like in the parallel traverser. that way we only need copies for maxConcurrency, not for all partitions, especially since we can't run all of them anyways.

knutwalker:
I changes the implementation to reuse bites data structures from previous runs on the same thread. This increased the runtime a bit has big improvements in memory consumption. The following improvements are from the GC Profiler output of the included JMH benchmarks on my machine.

| nodeCount | sourceCount |  time before |   time after | time diff | memory before | memory after | memory diff |
|----------:|------------:|-------------:|-------------:|----------:|--------------:|-------------:|------------:|
|      1024 |          32 |     0.797 ms |     1.080 ms | +35.508 % |     12.157 kB |    12.303 kB |    +1.201 % |
|      1024 |         128 |     1.026 ms |     1.202 ms | +17.154 % |     40.059 kB |    40.466 kB |    +1.016 % |
|      1024 |        1024 |     6.358 ms |     8.705 ms | +36.914 % |    317.961 kB |    84.504 kB |   -73.423 % |
|      1024 |        8192 |    49.324 ms |    65.667 ms | +33.134 % |   2541.722 kB |   154.922 kB |   -93.905 % |
|      8192 |          32 |    47.124 ms |    73.093 ms | +55.108 % |     96.176 kB |    96.333 kB |    +0.163 % |
|      8192 |         128 |    49.241 ms |    72.401 ms | +47.034 % |    308.861 kB |   309.281 kB |    +0.136 % |
|      8192 |        1024 |   419.792 ms |   566.204 ms | +34.877 % |   2468.958 kB |   625.407 kB |   -74.669 % |
|      8192 |        8192 |  3290.713 ms |  4521.812 ms | +37.411 % |  19749.200 kB |   692.919 kB |   -96.491 % |
|     16384 |          32 |   184.164 ms |   276.827 ms | +50.315 % |    192.229 kB |   192.409 kB |    +0.094 % |
|     16384 |         128 |   211.107 ms |   291.065 ms | +37.876 % |    616.133 kB |   616.525 kB |    +0.064 % |
|     16384 |        1024 |  1627.207 ms |  2415.381 ms | +48.437 % |   4926.830 kB |  1239.955 kB |   -74.833 % |
|     16384 |        8192 | 11694.258 ms | 16765.490 ms | +43.365 % |  39409.753 kB |  1307.075 kB |   -96.683 % |

benchmark/src/main/java/org/neo4j/graphalgo/bench/MSBFSBenchmark.java
```diff
@@ -0,0 +1,129 @@
...
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.concurrent.TimeUnit;
+
+@Threads(1)
+@Fork(value = 1, jvmArgs = {"-Xms4G", "-Xmx4G"})
```
alexaverbuch: suggest running with `-XX:-UseSuperWord` (prevent JVM from using vector/SIMD operations) & `-XX:+UseSuperWord` (allow JVM to use vector/SIMD operations)

this will tell you if your code is written in a way that SIMD optimizations can be made (that is a major contribution of this paper/algorithm)

regarding SIMD & Java: http://prestodb.rocks/code/simd/

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/BiMultiBitSet32.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Arrays;
+
+/**
+ * Two 32-bit wide BitSets per node.
+ * This class represents the fixed-size bit field as described in [1].
+ * Where two bit sets are stored in the same cache line to improve locality
+ * on sequential access.
+ * It is equivalent to a {@code List<List<BitSet>>}, but implemented more
+ * efficiently as a sparse {@code long[]}.
+ * The {@code ω} parameter is fixed to 32, that is, only 32 sources can be
```
knutwalker: The long is two bit sets in one, each 32-bit wide. As access happens sequentially, we can load the required values in one operation.

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/MultiSourceBFS.java
```diff
@@ -0,0 +1,335 @@
...
+public final class MultiSourceBFS implements Runnable {
+
+    // how many sources can be traversed simultaneously
+    private static final int OMEGA = 32;
+
+    private final ThreadLocal<MultiBitSet32> visits;
+    private final ThreadLocal<BiMultiBitSet32> nextAndSeens;
+
+    private final IdMapping nodeIds;
+    private final RelationshipIterator relationships;
```
knutwalker: RelationshipIterator is our Graph API, which might not always have an int[] backing it

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/MultiSourceBFS.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+
+        int depth = 0;
+
+        while (true) {
+            int nodeId = -1;
+            while ((nodeId = visit.nextSetNodeId(nodeId + 1)) >= 0) {
+                int nodeVisit = visit.get(nodeId);
+                assert nodeVisit != 0;
+                relationships.forEachRelationship(
```
knutwalker: Same as above, it's the trade-off of using our Graph API.

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/BiMultiBitSet32.java
knutwalker: I have written a little bit about using 64 or higher omega values in #206

benchmark/src/main/java/org/neo4j/graphalgo/bench/MSBFSBenchmark.java
knutwalker: thanks, will try it out!

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/BiMultiBitSet32.java
alexaverbuch: @knutwalker oh right, because you access (therefore store) `next` & `seen` together? (in `NextAndSeenLocal`)

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/MultiSourceBFS.java
alexaverbuch: this indirection might prevent the JIT & CPU from performing certain optimizations, e.g., prefetching.
maybe it's not a big deal, but I'd consider revising the API. maybe prototype an implementation of this algorithm where all arrays are arrays (rather than hidden behind iterators) to see if it's really costing you anything.

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/BiMultiBitSet32.java
knutwalker: @alexaverbuch yeah, this BiBitSet combines next & seen in a long since access to those is always sequentially for the same node id

algo/src/main/java/org/neo4j/graphalgo/impl/msbfs/MultiSourceBFS.java
knutwalker: Revising this API is certainly something outside of the scope of this PR though, but I can create an issue for this. With our current graph implementations, there's only the HeavyGraph that could provide in int[] without having to copy. The LightGraph might in a some cases, but it can't be guaranteed. So there's tradeoffs in those direction as well. OTOH, the graph internally does a simple loop over an array: https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/37769612b7657534ecc0eff71c45649f3d9743d3/core/src/main/java/org/neo4j/graphalgo/core/heavyweight/AdjacencyMatrix.java#L238-L244

PR #211, opened 2017-07-07
mutli source (parallel) closeness centrality

mknblch:
ClosenessCentrality implementation using MultiSource-BFS. 

The branch is based on the branch _asp-on-msbfs_


mknblch:
have to adapt some tests first

algo/src/main/java/org/neo4j/graphalgo/impl/MSClosenessCentrality.java
```diff
@@ -0,0 +1,88 @@
...
+        farness = new AtomicIntegerArray(nodeCount);
+    }
+
+    public MSClosenessCentrality compute() {
+
+        final BfsConsumer consumer = (nodeId, depth, sourceNodeIds) -> {
+            // TODO size of sourceNodeIds
+            while (sourceNodeIds.hasNext()) {
+                sourceNodeIds.next();
+                farness.addAndGet(nodeId, depth);
```
knutwalker: so, it's basically `farness.addAndGet(nodeId, depth * len(sourceNodeIds))`?

knutwalker: if so, we can add or modify the third parameter to return the size w/o having to iterate.

algo/src/main/java/org/neo4j/graphalgo/impl/MSBFSAllShortestPaths.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Iterator;
+import java.util.Spliterators;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * AllShortestPaths:
```
knutwalker: JavaDoc and Comments are still copy&pasted from the PD-ASP and need to be updated.

tests/src/test/java/org/neo4j/graphalgo/algo/ShortestPathDeltaSteppingProcTest.java
```diff
@@ -90,7 +90,7 @@ public static void setup() throws KernelException {
     @AfterClass
     public static void shutdownGraph() throws Exception {
         api.shutdown();
-//        Pools.DEFAULT.shutdownNow();
+//        Pools.DEFAULT.shutdownNow();y
```
knutwalker: most important change, probably 😉

algo/src/main/java/org/neo4j/graphalgo/impl/MSClosenessCentrality.java
mknblch: yes but can we have both :) ?

knutwalker: we can introduce a new interface/class to return there, should be no problem

tests/src/test/java/org/neo4j/graphalgo/algo/ShortestPathDeltaSteppingProcTest.java
mknblch: oops

PR #220, opened 2017-07-12
Various Enhancements

### LINKED ISSUE #29: Decide for one approach of Iterator vs Consumer vs Lookup in Graph API

### LINKED ISSUE #166: GraphView produces many failing tests
Running the test from #165 with the graph view as well produces may test errors, both as Exceptions during execution and as test assertion failures.

### LINKED ISSUE #199: Add Direction Filtering to Graph Loader
In order to be able to only load either OUTGOING or INCOMING relationships and not always having to load both, we want to add a filter via the GraphConfig/GraphLoader.

knutwalker:
- Remove all Iterator based APIs, decide for Consumer based APIs, Closes #29.
  - This one unhanded various test failures for the GraphView in the negative graph tests, since these were using the iterators (which worked) and now use the consumers (which were broken).
  - I refactored the GraphView consumer to turn the tests green, this might have also fixed #166 in the progress, but I haven't checked. (edit: it did, #166 can be closed as well)
- Ability to load only OUTGOING or INCOMING relationships, Closes #199.
  - If one direction is not loaded and an algorithm want's to access this direction, NPEs are thrown.
  - Those are not the best exceptions here, the alternative would be to add null checks on every forEach call
- Improved the representation of the Light Graph.
  - Use two separate adjacency arrays for in/out (allowing one of them to be null, if filtered)
  - encode degree as `offset[node + 1] - offset[node]`
    - we still need two array accesses for calculating the degree, but now they happen at least very likely in the same cache line instead of different ones.
  - add option to add multiple values in one go w/o page calculation
  - in my local PageRank LDBC benchmark, light and heavy were about the same
  - loading still only happens in a single thread, though


jexp:
@knutwalker currently the direction is hard coded in most procs as outgoing, I think it would be great if the user can supply the direction in the config array, which we would default to outgoing

@tomasonjo was also asking about that and also bi-directional loading, i.e. treating incoming and outgoing the same with regard to neighbors.

knutwalker:
@jexp direction is hardcoded rn because the algorithm implementations are using this particular direction only. If we have the direction configurable by the user, the algos must follow suit which may or may not change their semantics. LabelProp already does it this way, for the others we can decide whether variable direction makes sense (Also the WIP graph loader allows for user config).

Issue #218, opened 2017-07-11
Add progress logging


jexp:
for:

- [x] loading
- [x] algorithm execution
- [x] writing back

We should add some debug logging for these long running algos, 
either generically in ParallelUtil (e.g. a debug output every 10 seconds with some progress) or in the algos, we could have a central logger that keeps the last timestamp per "source" so that people see how far it has gotten.

Martin Knobloch can we get a logging instance at runtime without passing the Log object into each algorithm ?
and is it possible to activate log output using the embeddedDb in unitTests ?

Martin Knobloch because otherwise we would have to pass the Log from the procedure to each algorithm therefore we would have to adapt procs, alghorithm ctor signatures, tests

Martin Knobloch and i added some log output to one or two tests but didnt see anything in the tests

Michael Hunger to get the logger from the kernel via resolveDependency and pass it to ParallelUtil and the GraphFactory

Cleanest way would be:

- get log via @Context in procedure
- pass to graph factory and exporter via fluent API
- (while migrating) algorithm could use the graph for logging (via interface)
- later we can pass the Logger interface directly to each procedure

mknblch:
Some algorithms do not use the Graph interface but the api2 sources. The GraphLoader/Factoriy is not involved in every algorithm (see MST for example). And ParallelUtil is a bunch of static functions. Therefore we would have to do the refactoring anyway. 

In conclusion it is not possible for us to get the Logger through some injection magic.

My suggestion would be to create some kind of abstract class for algorithms which does not pretend any ctor signature but adds a NullObject-Logger (or sysout logger) by default and some fluent setters for it. This would safe us from adapting each unit-test ( until we test the logging itself ) but we could easily pass the logger object from the procedure to the algo instance.
The Exporter itself is already an abstract class so there is no problem.

mknblch:
Progress Logging is implemented in load, eval and write logic (where possible) using a interval of 10s for each log now. Due to the inaccuracy of the progress-calculation the values should be understood as approximation rather then exact values.

mknblch:
We should do some tests on a bigger graph for each algorithm. Im a bit concernded about the current interval of 10s which might be cool for the user if he's watching the log but it does polute log files if those algorithms run in a backend of a service or so. 

We could also think about lowering the LogLevel of the ProgressLogger or we could make interval and/or loglevel changable in the config.

mknblch:
implemented in loading, eval and write back

PR #221, opened 2017-07-13
add progress logging to algorithms

mknblch:
PR for Issue #218


knutwalker:
Build failure is unrelated to PR and fixed in #227

core/src/main/java/org/neo4j/graphalgo/core/utils/ProgressLoggerAdapter.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.function.DoubleSupplier;
+
+/**
+ * @author mknblch
+ */
+public class ProgressLoggerAdapter implements ProgressLogger {
+
+    private final Log log;
+
+    private int logInterval = 60_000;
```
jexp: I would make this configurable and reduce it to 10s by default, but can do it in a later change.

PR #228, opened 2017-07-14
Procedure for MultiSource-AllShortestPath

mknblch:
**UPDATE**

- i changed the behavior of `algo.allShortestPaths` so that MSBFS-ASP is used if no propertyName (weight property)  is given which calculates the unweighted shortest path (count of nodes). 
- otherwise - if a property is given - the parallel dijkstra ASP algo is used


knutwalker:
the shortest paths on MS-BFS couldn't calculate the distance based on weight b/c the MS-BFS doesn't (wouldn't and couldn't) emit the actual paths or accumulate their weights along the way. This ASP can only calculate the distance for equivalent to weight = 1

Issue #206, opened 2017-07-05
possible MS-BFS optimizations


knutwalker:
- return 1/0 from the next -> visit copy if any values have been copied so that we can see of visit is empty before iterating over it
- do a lazy reset of seen on the very first iteration while checking for the next set visitNext bit, saves the Array.fill before the first iteration
- support larger omegas: if nodeSize is < 2^28 we can support omega up to 256 (128 for 2^29, 64 for 2^30)
  - int[] as bitsets where every node has up to 8 consecutive ints as theirs bits, the second node starting at index 8 instead of index 1
- possibly double omega to [64, 512] if we remove the locality improved BiBitSet and use 3 BitSets with longs instead of ints.
  - that might also save us a copy step but we have to include the reset step again

PR #229, opened 2017-07-14
Remove Array fill operations in favour of batch initializing

knutwalker:
Part of the improvements mentioned in #206 (first two points). Playing around with different omega values would be very interesting, but also quite time consuming. However in the MS-BFS benchmark, we get good improvements already, drastic even for very large source sizes (before refers to c51dfce7bb41b3d7c9494cfda2a608e2daca7660).

<details>
<summary>benchmark diffs</summary>

| nodes | sources |  time before |   time after |   diff |
|------:|--------:|-------------:|-------------:|-------:|
|  1024 |      32 |     0.855 ms |     0.909 ms |  +6.3% |
|  1024 |     128 |     0.952 ms |     0.980 ms |  +2.9% |
|  1024 |    1024 |     6.900 ms |     2.122 ms | -69.2% |
|  1024 |    8192 |    55.036 ms |     5.461 ms | -90.1% |
|  8192 |      32 |    46.892 ms |    41.766 ms | -10.9% |
|  8192 |     128 |    50.793 ms |    51.129 ms |  +0.7% |
|  8192 |    1024 |   403.721 ms |   104.591 ms | -74.1% |
|  8192 |    8192 |  3227.843 ms |   130.017 ms | -96.0% |
| 16384 |      32 |   185.472 ms |   166.177 ms | -10.4% |
| 16384 |     128 |   180.980 ms |   182.295 ms |  +0.7% |
| 16384 |    1024 |  1442.495 ms |   365.722 ms | -74.6% |
| 16384 |    8192 | 11517.142 ms |   413.123 ms | -96.4% |

</details>


knutwalker:
A little explanation what changed and why this could have such an impact.

For every set of 32 sources (so for 8192 sources, thats 256 times) we traverse arrays of `nodeSize` length multiple times (the ones that are of interest here).

1. set initial bits for every source node (doesn't actually traverse the full length, but it's doing so with new arrays that aren't cached)
2. check if there is at least a bit set in some element
3. copy all visit bits to another array while resetting the seen bits
4. reset all bits to 0 for the next iteration

This PR merges those down to just two operations.

1. set initial bits for every source node while resetting all other its to 0.  This combines 1. and 4. from above, setting the source bits while the array is currently in a cache line, thus removing (basically guaranteed) cache misses
2. copy als visit bits while checking if at least one of them is set while resetting the seen bits. This combines 2. and 3., again, making use of the face that the array region we have to operate on is already cached.

Overall, this PR removes approx. `2 * SourceCount/32 * NodeCount / 16` (likely) dirty array accesses. For the last benchmark case, thats 0.5M array access operations that would be cache misses.

PR #231, opened 2017-07-17
[WIP] bfs & msbfs based connected components algorithms

mknblch:
fix benchmark & add MSColoring and single source coloring. both are connected component (union find) algorithms. 

the idea is basically to eval the color for each source-target node pair and assign the higher color to both. 

`org.neo4j.graphalgo.impl.MSColoring` adds this logic using `MultiSourceBfs` while `org.neo4j.graphalgo.impl.Coloring` is using`ParallelLocalQueueBFS`

also adds reset() method to MSBFS callback 

**TODO**

- add procedures
- add benchmark
- add test on more complex graph


algo/src/main/java/org/neo4j/graphalgo/impl/MSColoring.java
```diff
@@ -0,0 +1,109 @@
...
+                Runtime.getRuntime().availableProcessors() * 2, // TODO constant/getter anywhere!?
+                offset -> colors.set(offset, offset));
+    }
+
+    private void nodeAction(int node, int depth, BfsSources bfsSources) {
+        // evaluate highest color
+        int color = this.colors.get(node);
+        while (bfsSources.hasNext()) {
+            final int source = bfsSources.next();
+            color = Math.max(color, source);
```
mknblch: **TODO**
use color.get(source) instead of source

PR #234, opened 2017-07-18
Export progress

mknblch:
- moved the export logic from several classes into ParallelExporter
- created generic exporters usable for several algorithms (instead of one exp. for each algo)
- added ProgressLogging logic to ParallelExporter
- add progress logger testcase


tests/src/test/java/org/neo4j/graphalgo/algo/ShortestPathsProcTest.java
```diff
@@ -123,6 +123,8 @@ public void testResultStream() throws Exception {
             return true;
         });
 
+        System.out.println();
```
jexp: why do we need these?

core/src/main/java/org/neo4j/graphalgo/core/utils/ParallelExporter.java
```diff
@@ -1,33 +1,118 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
 
-public abstract class ParallelExporter<T> extends Exporter<T> {
+/**
+ * @author mknblch
+ */
+public abstract class ParallelExporter<T> extends AbstractExporter<T> {
 
-    private final ExecutorService executor;
+    public static final int DEFAULT_CONCURRENCY = Runtime.getRuntime().availableProcessors() * 2;
```
jexp: we should move this into some central place, 
also make concurrency configurable

core/src/main/java/org/neo4j/graphalgo/core/utils/ParallelExporter.java
```diff
@@ -1,33 +1,118 @@
...
-            ExecutorService executor) {
-        super(api);
-        this.batchNodes = batchNodes;
-        this.executor = executor;
-        this.batchSize = batchSize;
+    public static final String TASK_EXPORT = "EXPORT";
+
+    private final ExecutorService executorService;
+
+    private int concurrency = DEFAULT_CONCURRENCY;
```
jexp: take concurrency from procedure config via fluent setter

jexp:
Good call, I had some small comments regarding setting the concurrency, but I'll merge this anyway.

tests/src/test/java/org/neo4j/graphalgo/algo/ShortestPathsProcTest.java
mknblch: woops. forget to remove. i was testing load/eval/write progress log and needed a newline ;)

mknblch:
im going to fix them in a followup branch

PR #236, opened 2017-07-18
Documentation of pagerank

tomasonjo:
I have updated the pagerank docs to include the cypher loading queries. How do you feel about how i structured the queries in the docs ? Can this be an example of how to change other docs or do you have any better idea ? Need to also change the param syntax to include graph:"cypher", but I am not sure yet how should I proceed to not confuse the user.


jexp:
We should have a general section that discusses the import / export parameters and potential options, no need to repeat that in detail for each algorithm.

doc/minimumspanningtree.adoc
```diff
@@ -1,21 +0,0 @@
-= Minimum Spanning Tree
```
jexp: why did you remove this?

doc/pagerank.adoc
```diff
@@ -14,12 +16,16 @@ We assume page A has pages T1...Tn which point to it (i.e., are citations). The
 
 PR(A) = (1-d) + d (PR(T1)/C(T1) + ... + PR(Tn)/C(Tn))
 
-Note that the PageRanks form a probability distribution over web pages, so the sum of all web pages' PageRanks will be one.
```
jexp: do you rather want to move these more graph-theory notes that you removed to a dedicated section

doc/pagerank.adoc
```diff
@@ -48,20 +54,54 @@ CREATE (home)-[:LINKS]->(about)
...
 [source,cypher]
 ----
 
-CALL algo.pageRank.stream('Page', 'LINKS', {iterations:20, dampingFactor:0.85}) YIELD node, score
+CALL algo.pageRank.stream('Page', 'LINKS', {iterations:20, dampingFactor:0.85}) 
+YIELD node, score
 
 ----
 
-.Running algorithm and writing back results
```
jexp: why was this removed?

doc/pagerank.adoc
```diff
@@ -48,20 +54,54 @@ CREATE (home)-[:LINKS]->(about)
...
+.Loading the subset of the graph with cypher query ( can be used to create a virtual graph )
+
 [source,cypher]
 ----
+
+CALL algo.pageRank.stream('MATCH (p:Page) RETURN id(p) as id',
```
jexp: should always use LIMIT on the .stream procedures so that people don't accidentally run it on huge graphs.

readme.adoc
```diff
@@ -63,6 +63,23 @@ dbms.security.procedures.unrestricted=algo.*
 ----
 ====
 
+
+== Introduction
+
+Graph theory is the study of graphs, which are mathematical structures used to model pairwise relations between nodes. 
+A graph is made up of nodes (vertices) which are connected by relationships (edges). 
+A graph may be _undirected_, meaning that there is no distinction between the two nodes associated with each relationship, or its relationships may be _directed_ from one node to another.
+Relationships are what graph is all about: two nodes are joined by a relationship when they are related in a specified way.
```
jexp: add a newline before

doc/minimumspanningtree.adoc
tomasonjo: it was a duplicate, we already have minimum-weight-spanning-tree.adoc

doc/pagerank.adoc
tomasonjo: I am trying to find out how to tell people which of the four versions of (un)directed,(un)weighted graphs do our algos support. I think maybe the best way is to write a new section with each combination and a ticked checkbox of those that we support ?

Also remove too much information from the simple example.

doc/pagerank.adoc
tomasonjo: Actually I think they are a part of the original Google paper, so i`ll just add them back.

tomasonjo:
Made the simple example simple again :) added the versions and cypher loading sections. Later we can add which config to use for which version of the graph.

Issue #196, opened 2017-06-29
Abort Algorithms on Transaction Termination by User


jexp:
The user can ctrl-c in shell or click the X  in browser to terminate transactions.
We need to react to those, they show up here:

E.g. directly in ParallelUtil between batches or at the beginning / end of tasks.
We shouldn't check *too* often to not affect overall performance.

```
    public static boolean transactionIsTerminated(GraphDatabaseAPI db) {
        final KernelTransaction ktx = db.getDependencyResolver()
                .resolveDependency(ThreadToStatementContextBridge.class)
                .getKernelTransactionBoundToThisThread(true);

        Status reason = ktx.getReasonIfTerminated();
        if (reason != null) {
            return true;
        }

        return false;
    }
```

mknblch:
should we do this in a dedicated monitoring thread?

knutwalker:
Just found in docs: We can inject an `org.neo4j.procedure.TerminationGuard` with `@Context` into the procedures and use this one to check for terminated transactions (possibly only in the thread that is running the proc). The `check` method throws an `TransactionTerminatedException`.

jexp:
Not sure this is already in 3.1x, can you check?
Just the plain exception will not be user friendly, we might also have partially written back data.
But it would be a first step.

knutwalker:
It's not mentioned in the 3.1 docs, but the [whole segment](https://neo4j.com/docs/developer-manual/3.2/extending-neo4j/procedures/#injectable-resources) doesn't exists in 3.1. The class [exists in 3.1](https://github.com/neo4j/neo4j/blob/3.1/community/kernel/src/main/java/org/neo4j/kernel/impl/proc/TerminationGuardProvider.java#L51) though.
I wouldn't let the exception bubble to the user, but we can cancel with some meaningful message. Although, doesn't a terminated tx mean that the user has abandoned the tx and won't be able to see the error anyway?

If we check this after, say, every ParallelUtil.run, it might not stop every algorithm, since some are finished after the first round of threads has been fired. Checking within PU.run seems somewhat icky to me, related to mixed concerns and responsibilities. Checking in a separate monitoring thread 1) might not work for the TerminationGuard due to thread-stmt-bridge 2) opens up a whole new array of issues in how we want to abort running threads.

jexp:
Yes, aborting running threads would involve checking a volatile boolean flag somewhere, that causes the Thread/Runnable to exit early voluntarily.

[...]

PR #239, opened 2017-07-20
Terminate computation

mknblch:
First draft impl. of #196 

- add TerminationFlag iface which tells algorithms if its transaction has been closed
- TerminationFlagImpl is using a KernelTransaction to evaluate if its TX is gone
- implemented flag-check in algorithms
- i borrowed some code from the enterprise edition (seen in the testcase) to test killing the tx

still WIP - have to add some tests


algo/src/main/java/org/neo4j/graphalgo/AllShortestPathsProc.java
```diff
@@ -47,6 +51,8 @@
                     .load(configuration.getGraphImpl());
 
             return new MSBFSAllShortestPaths(graph, Pools.DEFAULT)
+                    .withLog(log)
+                    .withKernelTransaction(transaction)
```
jexp: it would be nice of the API of the Algos would be the TerminationFlag interface and we would do the wrapping in the proc.

jexp: e.g. with a static method on TerminationFlagImpl.wrap(ktx)

PR #253, opened 2017-08-04
release resources in graphs before write

mknblch:


jexp:
Would it make sense to have some nested release methods too, e.g. on IdMap, AdjacencyMatrix etc. So before setting them to null, they would also release their internal datastructures?

Issue #261, opened 2017-08-08
LPA does not overwrite writeProperty results


tomasonjo:
When I was trying out different iterations with the algo I noticed, that it does not overwrite property.
On neo4j 3.1.5 using algos 3.1.5.1.

Create data:

```
CREATE (nAlice:User {id:'Alice'})
,(nBridget:User {id:'Bridget'})
,(nCharles:User {id:'Charles'})
,(nDoug:User {id:'Doug'})
,(nMark:User {id:'Mark'})
,(nMichael:User {id:'Michael'})
CREATE (nAlice)-[:FOLLOW]->(nBridget)
,(nAlice)-[:FOLLOW]->(nCharles)
,(nMark)-[:FOLLOW]->(nDoug)
,(nBridget)-[:FOLLOW]->(nMichael)
,(nDoug)-[:FOLLOW]->(nMark)
,(nMichael)-[:FOLLOW]->(nAlice)
,(nAlice)-[:FOLLOW]->(nMichael)
,(nBridget)-[:FOLLOW]->(nAlice)
,(nMichael)-[:FOLLOW]->(nBridget)
,(nCharles)-[:FOLLOW]->(nDoug)
```

run the algo:

```
CALL algo.labelPropagation('User', 'FOLLOW','OUTGOING', {iterations:10,partitionProperty:'partition', write:true}) 
YIELD nodes, iterations, loadMillis, computeMillis, writeMillis, write, partitionProperty
```

Results1:

![lpa](https://user-images.githubusercontent.com/19948365/29093257-7e6ebcf8-7c89-11e7-8eb8-8df5d95d3036.png)

running algorithm for the second time... 

![lpa2](https://user-images.githubusercontent.com/19948365/29093309-b0ca9064-7c89-11e7-90bc-74b0f5c60caa.png)

Clearly something is wrong as there are no nodes loaded... I tried using different iterations to see if results still change with different iterations and they did not.

mknblch:
The `stats.nodes` actually is just the size of the result-map. It should be the same size as the count of nodes though. But it looks we have a bug in assigning the partition keys to the properties which results in a wrong node count value in the stats. Im currently working on it

mknblch:
fixed

PR #267, opened 2017-08-09
[WIP] LabelPropagation Fix 3.1

mknblch:
Tomas found a bug in LPA where not all nodes got a partition key. the PR should fix this


mknblch:
seems like there is another bug. maybe related to #261. In `LabelPropagationProcIntegrationTest` we create a Graph with 2 nodes but the nodeCount of the loaded Graph contains 12 Nodes. I have to investigate this further

PR #273, opened 2017-08-11
Fix Stackoverflow in MSBFS

### LINKED ISSUE #233: Likely Stackoverflow when running MS-BFS with 10k+ sources/nodes
#223 broke the MS-BFS for a large number sources. For example with 16384 sources, we create 512 runables and submit them at-once. Since only a handful can run at the same time and the queue for the pool is bounded, the majority of those tasks gets rejected and the rejection policy sleeps for a little while and then tries again. If there isn't a new thread available, the policy gets called recursively, consuming stack space. If there are enough nodes, rejections happens too often and the main thread runs in a SO exception.

Quick Solution: Use new pool with different rejection policy that block for a much longer time (difficult to tune) or runs the task itself or use an unbounded waiting queue for the pool. Ties MS-BFS closely to that pool, which is not a good thing.

Better Solution: Require concurrency parameter (which we want to do anyway), then submit at most `concurrency` tasks at once and wait for their completion before continuing. This could also be done in another method in ParallelUtil, which can submit new tasks once there is once new spot open.

knutwalker:
This is **not** based on #272, but has one conflict with it. I would rebase this one after 272 has been merged.

Introduces a new ParallelUtil method `runWithConcurrency`. It can be used to run an arbitrary collection of runnables on a given executor, but never more than `concurrency` at once, which is a parameter. Runnables that aren't running are not enqueued to the ExecutorService and depending on the collection implementation, they are not created until right before their execution. the MSBFS supplies a custom Collection implementation, which creates the tasks lazily and thus avoids flooding the ES queue into the stack overflow.
Furthermore, the new method supports the TerminationFlag since it can actually decide not to enqueue some tasks if the overall execution happens to be terminated, unlike the default run method, which would immediately enqueue all tasks and could only abort then by shutting the executor down.

Fixes #233.


knutwalker:
@jexp just rebased, ready for review

PR #292, opened 2017-08-22
Parallel triangle list

mknblch:


algo/src/main/java/org/neo4j/graphalgo/TriangleCountProc.java
```diff
@@ -0,0 +1,152 @@
...
+    public GraphDatabaseAPI api;
+
+    @Context
+    public Log log;
+
+    @Context
+    public KernelTransaction transaction;
+
+    @Procedure("algo.triangles.stream")
+    @Description("CALL algo.triangles.stream({nodeQuery:'labelName', relationshipQuery:'relationshipName', concurrency:8}) " +
```
jexp: can we move the label + rel-type to regular parameters like we have in the other procedures?

jexp:
@mknblch do you plan to add the procedure that returns the triangles in the graph (globally or per node) in a separate PR?

mknblch:
Sorry i did not mention that i replaced TriangleCount completely with TriangleList because this algorithm computes both, the per-node-count and the overall triangle count. If the stream() method is used, only the per-node-triangles are returned. in no-streaming-mode the sum of all triangles is returned. 

Note: the per-node-count counts each triangle 3 times at each node it connects

jexp:
This should go into the docs: Note: the per-node-count counts each triangle 3 times at each node it connects, can you add a docs page for it?

I think it would still be useful to have a procedure that returns the actual triangles as 3 columns with node-ids?

At some point we will have to look into de-duping. Not sure what the papers say.

mknblch:
i squashed all the changes into a new PR (#307)

PR #311, opened 2017-08-23
Add ExecService to remaining GraphLoader instances

mknblch:
unified and set ExecService to remaining GraphLoader instances


jexp:
@mknblch can we make them ask procedureConfig for the pool?

jexp:
Merging for now to prevent failures.

PR #317, opened 2017-08-24
Respect concurrency in heavy graph loader

knutwalker:
- light doesn't have parallel loading (yet)
- view doesn't need loading
- huge not yet included but will be adapted as well


jexp:
@knutwalker view needs loading for idmap-if we have ids with gaps.

knutwalker:
@jexp yes, it also does load an id map. It's just that this step doesn't run in parallel and the graph view has no loading step that could be run in parallel, so we don't need to set the proper concurrency wrt this PR.

PR #341, opened 2017-08-31
WIP clustering coefficient based on triangleCouning

mknblch:


mknblch:
unintentionally changed a var in ShortestPaths...

PR #345, opened 2017-09-01
Betweenness with direction

mknblch:
Add Direction parameter to Betweenness Algorithms. 

The Direction is given as optional argument in the configuration-map. 

E.g. `CALL algo.betweenness('','', {direction:'<>', concurrency:4, write:true, stats:true, writeProperty:'centrality'}) YIELD nodes, minCentrality, maxCentrality, sumCentrality, loadMillis, computeMillis, writeMillis`

**Following values are supported:**

Outgoing:
- `OUTGOING`
- `OUT`
- `O`
- `>`

Incoming:
- `INCOMING`
- `IN`
- `I`
- `<`

Both:
- `BOTH`
- `B`
- `<>`


ajkl:
Will this be ported to 3.2 ?

Issue #349, opened 2017-09-03
Betwenness centrality returns negative values 


tomasonjo:
I used the [facebook friend list](https://snap.stanford.edu/data/egonets-Facebook.html) (Edges from all egonets combined) to test out triangles and undirected betwenness. 

Import:

```
USING PERIODIC COMMIT 5000
LOAD CSV FROM "file:///facebook_combined.txt" as row fieldterminator ' '
MERGE (u:User{id:row[0]})
MERGE (u1:User{id:row[1]})
MERGE (u)-[:FRIEND]-(u1)
```

When i run betwenness centrality with any direction, all of the versions contain negative values.
I think that betwenness algorithm should not return negative values based on the definition.
```
CALL algo.betweenness('User', 'FRIEND', {direction:'both',write:true, stats:true, writeProperty:'centrality',concurrency: 4}) 
YIELD nodes, minCentrality, maxCentrality, sumCentrality, loadMillis, computeMillis, writeMillis
```

results: 

![negative](https://user-images.githubusercontent.com/19948365/30006382-999a6e6e-90f7-11e7-9c31-a549d2496847.png)

Some further checks:
```
MATCH (n:User) RETURN n.id,n.triangles,n.centrality ORDER by n.centrality asc limit 20
```
return 

![further](https://user-images.githubusercontent.com/19948365/30006421-4f9e052c-90f8-11e7-9e4b-e4a6eccd79c6.png)

mknblch:
this looks strange .. im going to check this

mknblch:
Update: I added a test to reproduce the error. It seems negative values only occur when using the Procedure. A test directly on the Graph led to correct values.

mknblch:
Update: only ParallelBC is affevted

mknblch:
The Problem is a silent integer overflow in the scaling logic of `org.neo4j.graphalgo.core.utils.AtomicDoubleArray`. When we set a value we multiply it by the `slaceFactor` (default is 100_000) and sum it up in the AtomicIntegerArray. During this routine an intereger overflow happens and negates the values.

mknblch:
Bug is adresses in #353

PR #353, opened 2017-09-05
IntegerOverflow fix in AtomicDoubleArray

mknblch:
This bug is responsible for #349


tests/src/test/java/org/neo4j/graphalgo/impl/BetweennessCentralityTest_349.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ *
+ * @author mknblch
+ */
+public class BetweennessCentralityTest_349 {
+
+    private static GraphDatabaseAPI db;
+    private static Graph graph;
+
+    private static final String PATH = "file:///Users/mknobloch/repository/neo4j-graph-algorithms/tests/src/test/resources/facebook_combined.txt";
```
knutwalker: doesn't really work on travis 😉

knutwalker: Ok, I should look at _all_ the changes first…

mknblch:
FYI: The testcase depends on local data and is therefore ignored

PR #355, opened 2017-09-05
fix for #349

mknblch:
The Problem was an integer overflow in the AtomicDoubleArray (because it scales doubles to ints). We solved it by capping possible overflows to Integer.MAX_VALUE. This means the maximum possible centrality value for a node is Integer.MAX_VALUE / scaleFactor (usualy 100_000). This must be documented.

I left the BigDoubleArray impl. in the project because we may need it in future.


tomasonjo:
```
This means the maximum possible centrality value for a node is Integer.MAX_VALUE / scaleFactor (usualy 100_000). 
This must be documented.
```

I can add this under implementation.
How is scale factor defined or what does it do ?

mknblch:
@tomasonjo We had to implement AtomicDoubleArrays on top of an AtomicIntegerArray. The scaleFactor is just the multiplicand to transform a double to int and back. Due to the scaling we get big values which may overflow the possible int-range. 

The scaleFactor is used in every algorithm that uses AtomicDoubleArray. It can be set with an optional scaleFactor parameter in the configuration map. Its default is 100_000.

PR #364, opened 2017-09-08
Track allocated memory

knutwalker:


core/src/main/java/org/neo4j/graphalgo/core/utils/paged/AllocationTracker.java
```diff
@@ -0,0 +1,65 @@
...
+            // allow for a bit of overflow before going to the next unit to
+            // show a diff between, say, 1.1 and 1.2 MiB as 1150 KiB vs 1250 KiB
+            if (bytes >> 13 == 0) {
+                return Long.toString(bytes) + unit;
+            }
+            bytes = bytes >> 10;
+        }
+        // we can never arrive here, longs are not large enough to
+        // represent > 8192 yobibytes
+        return null;
```
jexp: perhaps a proper error message instead :)

PR #379, opened 2017-09-18
Load nodes in parallel

knutwalker:
Implemented only for huge graph. It's not as parallel as we can since we're using just one thread to load all nodes, but it's a first throw at the problem.


jexp:
Parallel node-loading has so far no beneficial effect, the single threaded variant took 9-17 min last time, now up to > 60 minutes.

PR #383, opened 2017-09-19
Yelp intro docs

tomasonjo:
WIP
I wanted to show how I defined the structure of the docs for Yelp intro if you have any feedbacks. 
Semi-finished with introduction, gathering data about networks now so I can introduce them properly with some theory behind it.


tomasonjo:
Removed feature extraction as it might be confusing with ML feature extraction and dimensionality reduction. 
This are the three networks (social + review similairty + review co-occurence) I will use graph algos to demonstrate. 
Need to add stats about each network and then I think intro is done.

PR #396, opened 2017-09-27
Docs Concurrency update

tomasonjo:
I have found few things, that might be changed, so that as default we use as many threads as possible...

algo.scc.multistep uses:

```
configuration.getNumber("concurrency", 1).intValue()
```

algo.scc.forwardBackward uses:

```
 configuration.getConcurrency(1))
```

And correct me if I am wrong `algo.betweenness.exp1` seems to have hard-coded 4 threads:

```
 ParallelUtil.iterateParallel(executorService, nodeCount, 4, i -> {
            d.set(i, -1);
            sigma.set(i, 0);
            offsets.set(i, 0);
        });
```
when i compare it to:

```
ParallelUtil.iterateParallel(executorService, nodeCount, Pools.DEFAULT_CONCURRENCY, i -> {
                centrality.set(i, centrality.get(i) / 2.0);
            });
```

I can easily change `algo.scc.*` to default to that it will use maximum threads possible
```
configuration.getConcurrency()
```

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/AllShortestPathsProc.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/AllShortestPathsProc.java
@@ -33,7 +33,7 @@ public class AllShortestPathsProc {
 
     @Procedure("algo.allShortestPaths.stream")
     @Description("CALL algo.allShortestPaths.stream(weightProperty:String" +
-            "{nodeQuery:'labelName', relationshipQuery:'relationshipName', defaultValue:1.0}) " +
+            "{nodeQuery:'labelName', relationshipQuery:'relationshipName', defaultValue:1.0, concurrency:4}) " +
             "YIELD sourceNodeId, targetNodeId, distance - yields a stream of {sourceNodeId, targetNodeId, distance}")
     public Stream<AllShortestPaths.Result> allShortestPathsStream(
             @Name(value = "propertyName") String propertyName,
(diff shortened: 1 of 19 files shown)


jexp:
@knutwalker @mknblch ping.

jexp:
@mknblch was there a reason why we default to 1 for these procedures? Might be quite confusing for the users? Wouldn't it be more consistent to use the same concurrency approach everywhere?

jexp:
Also let's change the documented concurrency to 4 not 8. Most people  don't have that many cores on their machines.

tomasonjo:
I talked to @mknblch and he said there is no problem if I change the default for `algo.scc.*` to default `configuration.getConcurrency()`. There remains only betweenness then which runs on 1 thread as default.

tomasonjo:
updated

mknblch:
We should use `Pools.DEFAULT_CONCURRENCY` or just `conf.getConcurrency()` without argument in the procedures itself. In Descriptions i would suggest 4. 

If there are hardcoded values somewhere it's probably a leftover of some tests.

tomasonjo:
I also found `configuration.getInt("concurrency", 4)` in `algo.shortestPath.deltaStepping.stream`, that I changed to `conf.getConcurrency()`.

To recap what I changed:

Changed concurrency settings to use `conf.getConcurrency()`.for :
`
- algo.shortestPath.deltaStepping.stream
- algo.scc.multistep
- algo.scc.multistep.stream
- algo.scc.forwardBackward


Changed algo descriptions and docs to use `concurrency: 4`

PR #385, opened 2017-09-24
PageRank Performance Optimisations

pstutz:
Hi @jexp, @mknblch, and @knutwalker,

Thanks, that's a really nice project. I really like the chosen trade off between performance and access to abstractions, such as hiding the graph representation.

I implemented some optimisations for PageRank, are you interested in these changes? If you like them I'd be happy to update the comments. I think the precision changes in tests were because in this implementation the ranks grow upwards from `alpha`, whilst in the current implementation they approach the final rank from `1.0/numberOfNodes`. If you care for this I can run the test to full convergence and choose appropriate test deltas relative to the converged values.

Implemented optimisations:
- Delta PR: Only the rank delta is propagated instead of the whole rank. This allows to avoid work when the rank delta drops to 0. In the view graph representation this is especially valuable, as retrieving the degree for a node is expensive.
- Delta threshold: There is a new parameter that allows to already cut rank propagation short when the rank delta drops below a threshold. Set the default to `0.001`.
- `LightGraph.forEachRelationship` currently allocates a new cursor for every relationship traversal. This change avoids this with a new `Graph.singleThreaded()` method that allows to retrieve a view of a graph that only needs to be safe for single-threaded access, which allows to reuse the cursors. Each PageRank partition uses such a view for traversals.
- Changed `ComputeStep.srcRank` from an array of integers of size 1, to just an integer for cheaper access.

Other things in this PR:
- Updated Neo4j and JMH dependencies.
- Commented out the MultistepSCC and ParallelUnionFind benchmarks, as they do not compile on the current 3.2 branch.

PageRank benchmarks in PR with all changes, run on my local machine:
```
Benchmark                  (graph)  (impl)  (iterations)  (parallelBatchSize)  Mode  Cnt   Score   Error  Units
PageRankBenchmark.run          N/A   LIGHT             5                  N/A  avgt    5  12.972 ± 1.504  us/op
PageRankBenchmark.run          N/A   LIGHT            20                  N/A  avgt    5  15.786 ± 1.597  us/op
PageRankBenchmark.run          N/A   LIGHT           100                  N/A  avgt    5  21.206 ± 2.647  us/op
PageRankBenchmark.run          N/A   HEAVY             5                  N/A  avgt    5  14.376 ± 1.638  us/op
PageRankBenchmark.run          N/A   HEAVY            20                  N/A  avgt    5  15.367 ± 1.736  us/op
[...]

PR #401, opened 2017-09-29
Fix concurrency bug in HugeGraph

knutwalker:
When multiple threads started multiple nested iterations it could happen that those iterations ran on shared cursors and that it would throw around NPEs, AIOOBEs or just produce wrong data.

The proposed solution is to not use implicit pooling and reuse but to explicitly request new iterator instances that are safe to use independently from another. This is more or less the same solution that @pstutz proposed in #385.


pstutz:
I have these changes ready for 3.1 at home.

pstutz:
The reason I didn't send a PR test is because they weren't an improvement on HugeGraph in the benchmarks. If the current implementation is not safe that would explain why. :)

jexp:
@pstutz so you can send in your third change now?

pstutz:
@jexp: I could, but I mainly have a similar same change for LightGraph as well. Not sure of the semantics now, does this mean that by default all graphs are only safe for single-threaded access and one needs a copy for each thread that concurrently iterates to safely access it? I think it would be reasonable to make that the default, and it would allow for a simpler implementation than what I originally did with a separate extra class for each graph implementation that reuses its cursors.

knutwalker:
@pstutz I agree, I'd like to go that route to have on single-threaded access by default, makes it simpler and more explicit.

PR #402, opened 2017-09-29
Return long nodes in result stats if running with huge graph

knutwalker:


jexp:
Hmm, not too nice.

Could we do long in general? 

Or (Number) and number.longValue() the return object is required to hold long anyway.

knutwalker:
We can use just `long nodeCount` in id mapping, sure. Since Long#intValue does a possibly lossy (int) anyway, we can just use long und do the int casting ourselves, no need to go uppercase Long then.

jexp:
@knutwalker then let's do that.

PR #405, opened 2017-10-02
Several louvain fixes

mknblch:


tomasonjo:
It would be better if the default writeProperty was 'community' instead of 'cluster'... In my mind we want to use graph partitioning for weakly and strongly connected components and community detection for louvain and LPA and we avoid clustering

PR #421, opened 2017-10-13
Fix test gdb tests

jexp:


jexp:
Still a lot of test-failures remaining but which are more related to actual algorithm / data issues than mechanical issues.

PR #443, opened 2017-10-25
Louvain refactoring

### LINKED ISSUE #444: RelationshipWeights in HeavyGraph buggy for undirected algorithms
In case of the louvain algorithm we have to distinguish between directed and undirected graphs. When treating the Graph as undirected (using Direction.BOTH) its acutally not possible to have weights for incoming relationships. In this case the weight of A->B should be equal to the weight of B->A while these combinations should carry different weights when treating the graph as directed (using Direction.OUTGOING).

mknblch:
Louvain refactoring. Does not fix #444


knutwalker:
Travis tests point to an issue with weight importing when using BOTH, so not just an travis-only issue. Test may pass locally b/c the test graph is generated randomly.

PR #453, opened 2017-11-07
Parallel louvain

mknblch:
Unweighted, directed, parallel modularity based community detection algo.


jexp:
@mknblch there is a conflict now, could you rebase?

PR #454, opened 2017-11-07
AtomicDoubleArray use Long instead of integer

mknblch:


jexp:
Did you check on the memory increase? Is there any way to determine if we *need* a double size precision.

mknblch:
no i forgot about that. im going to check it tomorrow. 

I changed it to long since double is 64bit. We could either change its name to FloatArray or - as i did - change its underlaying datatype. This also allows another refactoring. See #455

PR #470, opened 2017-11-10
Hamonic Centrality 

mknblch:
The PR does *not* change closeness centrality but introduces another centrality based algorithm which should lead to the same distribution but with other absolute values. It does not suffer from disconnected graphs.

Its defined as the sum of the inverse of the shortest paths to each node while standard closeness centrality is defined by the inverse of the sum of distances. 

More [here](https://infoscience.epfl.ch/record/200525/files/[EN]ASNA09.pdf?)


algo/src/main/java/org/neo4j/graphalgo/HarmonicCentralityProc.java
```diff
@@ -0,0 +1,143 @@
...
+        final ClosenessCentralityProcResult.Builder builder = ClosenessCentralityProcResult.builder();
+
+        AllocationTracker tracker = AllocationTracker.create();
+        int concurrency = configuration.getConcurrency();
+        TerminationFlag terminationFlag = TerminationFlag.wrap(transaction);
+
+        Graph graph;
+        try (ProgressTimer timer = builder.timeLoad()) {
+            graph = new GraphLoader(api, Pools.DEFAULT)
+                    .withLog(log)
```
jexp: init()

jexp:
@tomasonjo see this update from @mknblch

jexp:
@mknblch NPE in test

- graph is loaded as outgoing
- MSHarmonic does forEachRel for BOTH
- if I change either of them the NPE goes away but then the values don't match
- please fix

```
Caused by: java.lang.NullPointerException
	at org.neo4j.graphalgo.core.heavyweight.AdjacencyMatrix.forEachIncoming(AdjacencyMatrix.java:274)
	at org.neo4j.graphalgo.core.heavyweight.AdjacencyMatrix.forEach(AdjacencyMatrix.java:206)
	at org.neo4j.graphalgo.core.heavyweight.HeavyGraph.forEachRelationship(HeavyGraph.java:84)
	at org.neo4j.graphalgo.impl.msbfs.MultiSourceBFS.run(MultiSourceBFS.java:204)
```

PR #478, opened 2017-11-20
Optimized triangle counting

pstutz:
An optimized parallel triangle counting implementation:
- Doesn't use the graph abstraction, but iterates directly over the adjacency arrays
- Uses F/J to parallelise
- When computing the intersection it scans over the smaller array and checks for existence in the other one with binary search
- Uses LArray to support arrays with more than integer max-value entries

Benchmark on laptop. This benchmark has only very tiny adjacency arrays and ~1000 nodes, so both binary search and F/J parallelisation don't pay off for it. This means that most of the improvements in this benchmark probably result from skipping the graph iteration abstractions, from not doing the book-keeping for recording found triangles, and from not checking for early termination.
```
TriangleCountBenchmark._01_triangleCount                   1  avgt   10  167.153 ±  5.594  us/op
TriangleCountBenchmark._01_triangleCount                   2  avgt   10  158.133 ± 20.737  us/op
TriangleCountBenchmark._01_triangleCount                   4  avgt   10  130.576 ± 17.691  us/op
TriangleCountBenchmark._01_triangleCount                   8  avgt   10  119.110 ±  7.240  us/op
TriangleCountBenchmark._02_triangleCountExp                1  avgt   10  167.416 ±  2.638  us/op
TriangleCountBenchmark._02_triangleCountExp                2  avgt   10  155.895 ±  3.641  us/op
TriangleCountBenchmark._02_triangleCountExp                4  avgt   10  125.308 ±  1.139  us/op
TriangleCountBenchmark._02_triangleCountExp                8  avgt   10  125.195 ±  1.162  us/op
TriangleCountBenchmark._03_triangleCountOptimized          1  avgt   10   23.259 ±  0.208  us/op
```


pstutz:
Added the functionality to return and iterate over all found triangles (time below includes constructing, returning, iterating over, and counting all triangles):
```
TriangleCountBenchmark._03_triangleFindOptimized          8  avgt   10  76.520 ± 2.873  us/op
```

mknblch:
FYI: We are about to implement an Interface, for checking if an relationship exists in the Graph API.

pstutz:
Great, I guess in that case the LArray[Array[Long]] with fast binary search `existsRel` checks on the sorted arrays could be an alternative `HugeGraph` representation. Did you benchmark the cost of using the iterators instead of accessing arrays directly? In some cases it might make sense for an algorithm to be able to rely on a specific graph implementation that allows direct access to the representation. Also I really like the Java F/J pool, did you benchmark it vs. the executor services? In my own experiments I found it almost impossible to match what F/J does and I guess its recursive variation might also offer a neat way to terminate an algorithm early by shutting down the pool instead of relying on constant checks from within the algorithm implementation.

mknblch:
I think we already have an (unsolved) issue regarding the use of arrays instead of our API though not with LArray youve been using. @knutwalker evaluated LArray as a possible candidate for the Huge Graph I guess but it seems it didnt went so well. And yes i think we could benefit from having direct access to the arrays instead of using the API but yet we haven't had the resources to evaluate and implement it probably.  If you like the idea you could create a new GraphInterface which gives access to the underlying arrays and implement a DataSource accordingly. I did that for a CC and MST algo too but its hard to maintain them when upgrading to another Neo4J Version.

Anyway. We have tested some different approaches to the F/J Pool in the experimental UnionFind implementations. @knutwalker evaluated the use of a queue to be the fastest. It basically puts all results on a queue and then merges them pairwise once two are available until only one remains. Although I would like to have one for the implementation of Algos :) The problem is that we dont have a custom F/J implementation.

pstutz:
@knutwalker: I'd love to learn more about the LArray benchmarks if you still have the results.

@mknblch: I imagine the maintenance for the custom graph implementations could be reasonable if they can be built from a Neo4j data source that is kept stable across Neo4j versions, but I haven't looked into that part of the codebase yet.

The split and merge approach you outline for UnionFind is essentially what I did here with the recursive F/J task. If you managed to substantially beat the performance of a recursive F/J task on this then I'd love to learn more about it. I tried many times to do better than F/J with custom schedulers and always failed. What would be the advantage of a custom F/J implementation over the regular one?

mknblch:
As far as i know there was a problem with concurrent writes on the LArray which we need during the import. 

Regarding the data source, though it would be nice to have, it's not likely that we will support one since only the HeavyGraph could implement it. The Huge and Light Graphs do a special encoding on the data in its arrays (VInt or interlacing with offsets etc.) So you cant treat them as container for nodeIds nor sort or change them without destroying its integrity. And we are currently about to adapt the algos to the Huge Graph (on top of the bugfixing) therefore algorithm optimization is out of our scope yet.

But Im going to talk with @jexp about the F/J Pool

Greets

mknblch:
So to make your algorithm compatible with our api you could either implement your own dataSource or let the algo use HeavyGraph. See #480 this one will add an existence check for relationships. If you cant do that without direct access to the Arrays you could hack your way to the `org.neo4j.graphalgo.core.heavyweight.AdjacencyMatrix` which is basically the container for incoming and outgoing relationships, their offsets and their degrees.

jexp:
For huge graph we could store with each segment the min and max id's its containing.

And filter out segments that are outside the range.
And linearly scan those that are within the range of the ids.

Perhaps it would be good to have a bulk operation for connectedness check, so we can more efficiently search for a bunch of target ids in bulk.

graph.areConnected(sourceId, long[] targetIds) -> boolean[] or perhpas long[] of the successful target-ids.

(or the same with streams)

We could also have a parallel post-load step that reorganizes segments into subsequent id's.

PR #479, opened 2017-11-20
Random Approximate Brandes BC - Fix

mknblch:
Fixes some bottlenecks.

**MaxDepthBrandesBetweennessCentrality** has been removed in favor of **RABrandesBC** since it also supports a maximum traversal depth parameter which is Integer.MAX by default. 

Preallocated arrays (sigma, delta) of length **nodeCount** for each Thread has been changed to HPPC structures which should reduce the amout of memory needed drasticaly.


jexp:
@mknblch can you also add / update docs for this one?

PR #497, opened 2017-12-05
[WIP] binary search refactoring

mknblch:


jexp:
Good pulling it out.

Still misses the linearSearch for sorted arrays that's called from binarySearch.
And which aborts as soon as the value is >= the key.

PR #506, opened 2017-12-05
Pull node ids one by one from queue for loading huge relationships

knutwalker:


knutwalker:
This is not compatible with #505, I can rebase either of the PR after the other's been merged though.

PR #511, opened 2017-12-10
Fix MST

mknblch:
- fix prim impl. 
- changed the signature: move label, relType and weightProperty into function instead of config
- removed sum-, max- and min-weight from result type (because O(n) )
- added a new field `effectiveNodeCount` which basically counts the visited nodes. This might come in handy for validation if the mst is calculated on a disconnected graph. 
- changed procedure name to `algo.spanningTree` /  `algo.spanningTree.minimum`
- the algorithm now also computes the maximumSpanningTree with  `algo.spanningTree.maximum` 
- both algorithms write a new relationship type (which can be set using the optional config param `writeProperty`) to the graph using the same weightProperty
- moved kSpanningTree into different branch due to bugs


mknblch:
**TODO**
- add min-max 
- assign original weight to exported relationship

PR #519, opened 2017-12-14
Added means to load named graphs to be repeatedly used by procedures later

jexp:


knutwalker:
LGTM

jexp:
Merged Manually.

PR #525, opened 2018-01-05
huge strongly connected components

mknblch:


mknblch:
Back to WIP. Forgot a procedure for Huge(recursive)Tarjan

mknblch:
I removed HugeTarjan (the recursive version) because it needs datastructures we do not have yet (Objects Bags for huge). This means only `algo.scc.*` & `algo.scc.iterative.*` work on huge datasets. But since Iterative Tarjan is already faster than the recursive version it should be fine

PR #527, opened 2018-01-05
Betweenness centrality

mneedham:


doc/betweenness-centrality.adoc
```diff
@@ -1,47 +1,57 @@
...
-over the flow throughout the full graph. For example, Alice in above example is the
-main connection in the graph. Were Alice to be removed, all connections in the graph would be cut off. This makes Alice “important”, because it ensures that no nodes are isolated. When using
-betweenness centrality as an analysis measure, it indicates a potential gate keeping or controlling node. 
+In the above example Alice is the main connection in the graph.
+If Alice is removed all connections in the graph would be cut off.
+This makes Alice “important” because she ensures that no nodes are isolated.
 
-It differs from the other centrality measures. A node can have quite low degree, be connected to others that have low degree, even be a long way from others on average, and still have high betweenness. Consider a node A that lies on a bridge between two groups of vertices within a network. Since any path between nodes in different groups must go through this bridge, node A acquires high betweenness even though it is not well connected (it lies at the periphery of both groups).
+The betweenness score for a node can be quite deceptive.
```
tomasonjo: I wouldn't call it deceptive. Maybe a different view of the importance of nodes from other centralities.

mneedham: Fair enough. How about if we say it's 'surprising' or 'not intuitive'?

tomasonjo: Not intuitive would be better I think

PR #528, opened 2018-01-07
Docs update k-mst

tomasonjo:
@mknblch 
For K-MST I have changed partitionProperty to writeProperty to be consistent with MST algo.
I have noticed that we can't actually get the k-mst tree back, but only its nodes and then deduce which relationships have been used. I have no idea if people need those relationships for their use-cases.

@mneedham 
I have changed the 
```
:img: https://github.com/neo4j-contrib/neo4j-graph-algorithms/raw/3.1/doc/img
to 
:img: https://github.com/neo4j-contrib/neo4j-graph-algorithms/raw/3.2/doc/img
```
Not sure if this breaks anything ?


doc/minimum-weight-spanning-tree.adoc
```diff
@@ -75,17 +152,50 @@ image::{img}/mst_result.png[]
 .Running algorithm and writing back results
 [source,cypher]
 ----
-CALL algo.spanningTree(label:String, relationshipType:String, weightProperty:String, startNodeId:long, {writeProperty:String}) YIELD loadMillis, computeMillis, writeMillis, weightSum, effectiveNodeCount
+CALL algo.spanningTree(label:String, relationshipType:String, weightProperty:String, startNodeId:int, {writeProperty:String})
```
mneedham: Is this an accidental change? In Java land this number is a `long` so I think it makes sense to use those types in the docs.

tomasonjo: We were using only int and float in docs till now so we do not differentiate between int and long as for end user it shouldn't make any difference

mknblch:
@tomasonjo: kMin/kMax ist just a shortcut for:
- build MST
- traverse and collect significant edges
- cut the tree at those edges
- assign unique id to each resulting branch
- export

If one needs the MST itself he can do the procedure by hand with cypher or run MST again on one of the resulting branches.

PR #529, opened 2018-01-11
Cypher projection 

tomasonjo:


readme.adoc
```diff
@@ -135,6 +159,19 @@ CALL algo.pageRank(
 {graph:'cypher', iterations:5, write: true});
 ----
 
+Cypher projection can also be used to project a virtual (non-stored) graph. 
+Here is an example how to project a stored bi-partite graph to a virtual mono-partite graph.
```
mneedham: Can we change this bit so that it doesn't use such technical language? Otherwise the reader will need to lookup what a 'bi-partitite' and 'mono-partite' graph is. 

In this case we can say:

"Here is an example of how to project a graph of people who visited the same web page" 

Or something along those lines - maybe there's a better way to phrase that

tomasonjo:
added also

```
Relationships described in the relationship-statement will only be projected if both source and target nodes are described in the node-statement.
All other relationships that don't have both source and target nodes described in node-statement will be ignored in projection
```

Issue #464, opened 2017-11-09
Use defaults weights/property values in cypher loading


knutwalker:
After implementing `cypher` for LPA, @tomasonjo found that the weight and value properties aren't optional: https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/463.

All other loaders use the default weight or default property value if a node or relationship doesn't have to configured one defined, whereas the cypher loaders fails. It should behave like the other loaders and use the default values in case the weight/value property wasn't even defined in the cypher statements.

mneedham:
Fixed by: https://github.com/neo4j-contrib/neo4j-graph-algorithms/commit/7337e65062ba78c76e1cb6b7e50099c851e22fb5

PR #531, opened 2018-01-15
Default weight/property loading from Cypher projections as per #464

mneedham:
This is pretty much how the other loaders work except the throwing of the `property doesn't exist` exception is thrown further down the stack and they only see a `null` value from the API they call.

### DIFF
--- a/core/src/main/java/org/neo4j/graphalgo/core/heavyweight/HeavyCypherGraphFactory.java
+++ b/core/src/main/java/org/neo4j/graphalgo/core/heavyweight/HeavyCypherGraphFactory.java
@@ -279,9 +279,9 @@ public class HeavyCypherGraphFactory extends GraphFactory {
                 }
                 if (hasRelationshipWeights) {
                     long relId = RawValues.combineIntInt(source, target);
-                    Object weight = getProperty( row, "weight" );
-                    if ( weight != null ) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/neo4j/graphalgo/core/heavyweight/HeavyCypherGraphFactory.java
```diff
@@ -280,15 +279,18 @@ public boolean visit(Result.ResultRow row) throws RuntimeException {
                 }
                 if (hasRelationshipWeights) {
                     long relId = RawValues.combineIntInt(source, target);
-                    relWeigths.set(relId, row.get("weight"));
+                    Object weight = getProperty( row, "weight" );
```
knutwalker: Since we did not adopt the Neo4j style guide, we might end up in a couple of format-heavy PR change sets here or there.

knutwalker:
LGTM

core/src/main/java/org/neo4j/graphalgo/core/heavyweight/HeavyCypherGraphFactory.java
mneedham: Yeh sorry about that. I'll use the default Intellij formatting settings when working on this project so that everything stays consistent.

Issue #530, opened 2018-01-15
[Docs] Document how to filter infinite values 


mneedham:
Some algorithms returning `infinity` and there isn't an easy way to filter these results using pure Cypher.

https://github.com/neo4j-contrib/neo4j-graph-algorithms/pull/441 provides some functions that we can use. 

Let's add an example showing how to use them.

Once this one is done we can close https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/427 and https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/256

PR #536, opened 2018-01-22
Filtering infinity docs

tomasonjo:
fix https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/530


mneedham:
@tomasonjo  This LGTM so we'll merge it.

But I think there are other algorithms that can also return infinite values when they try to calculate the distance to an unreachable node. e.g. closeness centrality might have the same issue.

So I wonder if we should create a section that talks in general about infinite values and link to that section from each of the algorithms

tomasonjo:
I have tried to reproduce this issue on closeness centrality but failed. I will try again on a bigger graph to check if I can reproduce it. 

Closeness centrality has a different issue on a disconnected graph, will provide a testcase and expected results tomorrow.  

If only SSSP and possibly closeness has infinity issue I wouldn't create a separate section for it.

PR #541, opened 2018-01-24
HugeTriangleCount

mknblch:


mknblch:
- adapt trianglecount to work with >2bn nodes

PR #543, opened 2018-01-25
Harmonic

mneedham:


doc/harmonic-centrality.adoc
```diff
@@ -1,47 +1,50 @@
 = Harmonic Centrality
 
 // tag::introduction[]
-The harmonic mean has been known since the time of Pythagoras and Plato as the mean expressing “harmonious and tuneful ratios”, and later has been employed by musicians to formalize the diatonic scale, and by architects as paradigm for beautiful proportions.[1]
+The Closeness Centrality algorithm is very sensitive to a single large distance or missing relationship - the closeness centrality for a node is 0 if there is just one other node it can't reach.
+If there's one node that no other nodes can reach then every node will have a closeness centrality of 0.
```
tomasonjo: Not entirely sure if this is correct... We are currently returning the opposite, check https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/547. As far as I understand for an undirected graph the way to solve above problem is to run closeness centrality on each single connected component, while for directed graph I am not sure what the best approach is when there are pairs of un-reachable nodes within a single CC...

mneedham: @tomasonjo hmm interesting. I'd got that understanding from:



in this paper that's referenced lower down the page - https://infoscience.epfl.ch/record/200525/files/[EN]ASNA09.pdf?version=1

tomasonjo: This is an example for undirected graphs, where nodes are unreachable only if they don't belong in the same connected component. We don't follow their logic for now... see https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/546

As for directed graphs I haven't found any useful articles yet...I figure that having pairs of unreachable nodes within a single component of a directed network can be more frequent and so more of a problem. If nodes that can't reach all the other nodes in the graph should have closeness value 0, they don't for now in our implementation.

mneedham: I'm gonna change that text to talk about this instead:

> When a graph is not strongly connected, a widespread idea is that of using the sum of reciprocal of distances, instead of the reciprocal of the sum of distances, with the convention {\displaystyle 1/\infty =0} 1/\infty=0:

> {\displaystyle H(x)=\sum _{y\neq x}{\frac {1}{d(y,x)}}.}

From https://en.wikipedia.org/wiki/Closeness_centrality - that's what we've seen from the few experimental graphs on #547

mneedham: @tomasonjo ok I pushed a new description for that bit.

tomasonjo:
LGTM

PR #565, opened 2018-02-09
A* Star shortest path algorithm implementation

smadurange:
Added procedure, implementation and unit test for a* algorithm for geospatial shortest path searches based on longitude and latitude.


mneedham:
@smadurange you need to build this against the 3.2 branch because we're cherry-picking commits from 3.2 ->3.3 -> 3.4 as there are some internal Neo4j API changes in between branches. Unfortunately that means that the trees don't match as you can see by the massive number of commits your PR has tried to pull in!

smadurange:
@mneedham , oh that is what you meant by sent it against 3.2. Sorry, I misunderstood. I will build it against 3.2 and send a new PR. Thanks.

PR #568, opened 2018-02-14
Parallel weightd louvain & deadlock fix

mknblch:
our current impl. of weighted louvain utilizes multithreading only during initialization. computation is done sequentialy. this implementation performs a randomized sequential modularity calculation on each thread and takes the best candidate after each run until no improvement can be found.


mknblch:
- enable huge & heavy graph in testcase
- fix in shuffledNodeIterator

PR #573, opened 2018-03-07
Closeness update

mknblch:
as usual
```
Tests with failures [seed: 39BC16839C8F5690]:
  - org.neo4j.graphalgo.core.utils.ParallelUtilTest.shouldRunAtMostConcurrencyTasks
  - org.neo4j.graphalgo.core.utils.ParallelUtilTest.shouldSubmitAtMostConcurrencyTasksRunSequentially
```


mneedham:
@mknblch are we intentionally now always treating graphs that use closeness as undirected?

mneedham:
@mknblch only other thing is I wonder if we should default to having the normalized/wassermanFaust option set to true

PR #578, opened 2018-03-13
More HugeLongArray improvements

knutwalker:
Includes #577


mneedham:
@knutwalker maybe I made a mistake here...I merged #577 - should I have actually merged this instead?

knutwalker:
@mneedham No, it's ok. I was working on top of #577 and probably got the commit history mixed up, I'll fix the PR.

knutwalker:
@mneedham PR is all fixed now

PR #579, opened 2018-03-14
Change MS-BFS to search 64 sources concurrently instead of 32

knutwalker:
The number of sources that can be traversed concurrently is limited by the number of bits that we can pack in a single per-node identifier. We need three of those per-node bitsets for MS-BFS. Until this PR, we used `int[]` arrays for those bitsets and had an optimization for memory locality where two bitsets were interleaved into a single `long[]` array where each long contained the 32-bit wise bitmap for a single node.

This PR changes this to use three separate `long[]` arrays, which allows us to traverse 64 source concurrently. The performance gains for traversing more sources outweigh the gains from the memory optimization.
Furthermore, the original paper goes up to 512 bits by using wide registers of AVX-512 compatible CPUs, which unfortunately cannot be exploited in Java, at least not until some form of [valhalla](http://jdk.java.net/valhalla/) lands.

Here are the improvements in runtime performance and heap usage for this PR (the parameter mean <number of total nodes in graph>_<number of sources to traverse from>)

msbfs-before-after-comparison

The concrete numbers for this PR are

msbfs-updated


mneedham:
@knutwalker I've merged in #578 now so this one needs rebasing

knutwalker:
@mneedham done

PR #584, opened 2018-03-18
Removed hard-coded references to latitude and longitude in A* implementation

smadurange:
1. Refactored proc and implementation so that the client can pass property key names for latitude and longitude
2. fCosts was not being reset properly. Fixed.


mneedham:
LGTM

PR #600, opened 2018-04-13
Fixing SSSP/D for cases where a node is queued multiple times before being traversed.

### LINKED ISSUE #599: SSSP returns incorrect result
I am working to implement the [LDBC Graphalytics](http://graphalytics.org/site/dist/index.html) workload using the Graph algorithms library.

I ran into an issue with SSSP. When looking for the shortest path between vertices 1 and 4, the algorithm finds the path with a cost of 5 (1->4), instead of the path with a cost of 4 (1->2->4->6->3->4).

The Cypher commands are as follows:

```
CREATE
  (v1:Node {VID: 1}),
  (v2:Node {VID: 2}),
  (v3:Node {VID: 3}),
  (v4:Node {VID: 4}),
  (v5:Node {VID: 5}),
  (v6:Node {VID: 6}),
  (v7:Node {VID: 7}) 
  ,(v1)-[:EDGE {WEIGHT: 0.5}]->(v2)
  ,(v1)-[:EDGE {WEIGHT: 5.0}]->(v3)
  ,(v2)-[:EDGE {WEIGHT: 0.5}]->(v5)
  ,(v3)-[:EDGE {WEIGHT: 2.0}]->(v4)
  ,(v5)-[:EDGE {WEIGHT: 0.5}]->(v6)
  ,(v6)-[:EDGE {WEIGHT: 0.5}]->(v3)
  ,(v6)-[:EDGE {WEIGHT: 4.0}]->(v7)
  ,(v1)-[:EDGE {WEIGHT: 5.0}]->(v4)
```

```
MATCH (startNode {VID: 1}), (endNode {VID: 4})
CALL algo.shortestPath(startNode, endNode, 'WEIGHT', {direction: 'OUTGOING'})
YIELD nodeCount, totalCost, loadMillis, evalMillis, writeMillis
RETURN totalCost
```

image

The issue is not trivial to reproduce, however, I believe, my [example project](https://github.com/szarnyasg/sssp-issue) manages to consistently reproduce the issue.

* Removing the 6->7 edge (which has a weight of 23 and has nothing to do with either of the paths) resolves the issue, see https://github.com/szarnyasg/sssp-issue/blob/949cc8e8ed9772fb23029bc9a1e9fa2b21139aa0/src/main/java/sssp/SsspMain.java#L31
* Changing the order of the edge insertions can also resolve the issue.
* I tried running these queries repeatedly in a Neo4j server. First, I used the following approach: 1) delete all nodes in the database 2) insert the data 3) invoke the SSSP procedure again. This failed to reproduce the issue consistently. So I changed my approach: 1) stop the database, delete the database directory, restart the database 2) etc. This managed to reproduce the issue consistently, which might give a hint on the nature of the problem (maybe it has something to do with internal edge ids?).

knutwalker:
Reported and based on the examples in #599, also fixes #599.

The basic issue is that the priority queue doesn't support duplicate values, but we might enqueue a node multiple times with different weights, coming from different paths.
The PQ is an array based heap and new nodes are enqueued by adding them to the end and then bubbling them up to their correct place.

First, we discover the target node (3) with a weight of 5 – it is places in the heap.
Then we discover the node (2) with a weight of 5. It's placed at the end of heap and doesn't bubble up, since its weight is not less than that of the previous node.
Then we discover a new path to (2) with a weight of 2. (2) is now placed at the end of the heap again and should bubble up, but the first check happens against the previously inserted node 2, for which the cost has changed as well.
Instead of moving in front of the node (3), the new (2) stays at the end (besides being twice in the queue) and the heap invariant is broken, as the smalled node is no longer at the front.
On the next queue.pop, we get (3) instead of (2) and think that we are finished with the path finding.
If we had discovered (2) first and (3) later, it would have worked by accident; the order of relationship traversal during the import matters here, as it defines whether we check (2) or (3) first.

To fix this, I added an update method to the PQ, which searches for an element in the queue and moves it to its proper place.
If we change the current cost leading up to a node, that we update the queue instead of adding the node again.

I also patched up A* and ShortesPath_s_, as well as AllShortesPaths, although that one would have eventually found all shortest paths, even if the queue was in a broken and suboptimal state during the algo run.
I've also thrown in a fix for the bitset, so that it doesn't copy its buffer on every put.

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/impl/ShortestPathAStar.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/impl/ShortestPathAStar.java
@@ -1,8 +1,10 @@
 package org.neo4j.graphalgo.impl;
 
-import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
-
+import com.carrotsearch.hppc.IntArrayDeque;
+import com.carrotsearch.hppc.IntDoubleMap;
+import com.carrotsearch.hppc.IntDoubleScatterMap;
+import com.carrotsearch.hppc.IntIntMap;
+import com.carrotsearch.hppc.IntIntScatterMap;
 import org.neo4j.graphalgo.api.Graph;
 import org.neo4j.graphalgo.core.utils.ProgressLogger;
 import org.neo4j.graphalgo.core.utils.queue.IntPriorityQueue;
@@ -12,19 +14,16 @@ import org.neo4j.graphdb.Direction;
 import org.neo4j.graphdb.Node;
 import org.neo4j.kernel.internal.GraphDatabaseAPI;
 
-import com.carrotsearch.hppc.IntArrayDeque;
-import com.carrotsearch.hppc.IntDoubleMap;
-import com.carrotsearch.hppc.IntDoubleScatterMap;
-import com.carrotsearch.hppc.IntIntMap;
-import com.carrotsearch.hppc.IntIntScatterMap;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 public class ShortestPathAStar extends Algorithm<ShortestPathAStar> {
-	
-	private final GraphDatabaseAPI dbService;
-	private static final int PATH_END = -1;
-	
-	private Graph graph;
-	private final int nodeCount;
+
+    private final GraphDatabaseAPI dbService;
+    private static final int PATH_END = -1;
+
+    private Graph graph;
+    private final int nodeCount;
     private IntDoubleMap gCosts;
     private IntDoubleMap fCosts;
     private double totalCost;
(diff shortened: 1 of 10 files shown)


szarnyasg:
Many thanks for the quick bugfix! I tested this branch with my project and it indeed resolved my issue.

algo/src/main/java/org/neo4j/graphalgo/impl/ShortestPathAStar.java
```diff
@@ -87,50 +95,56 @@ private void run(final int goalNodeId, final String propertyKeyLat, final String
...
+        return distance;
     }
-    
+
     private double getNodeCoordinate(final int nodeId, final String coordinateType) {
-    		final long neo4jId = graph.toOriginalNodeId(nodeId);
-    		final Node node = dbService.getNodeById(neo4jId);
-    		return (double) node.getProperty(coordinateType);
+        final long neo4jId = graph.toOriginalNodeId(nodeId);
+        final Node node = dbService.getNodeById(neo4jId);
```
jexp: @mknblch should we really use non-kernel API here?
And why didn't we load this data initially with the graph?

mknblch: I guess thats my fault. A* was a contribution from the community and since our property-loading-mechanism cannot load multiple properties per node I told the author to use the kernel api instead to get it done. Im going to talk with paul & mark about a change in the Weights-interface. I have some ideas how to adapt our current impl.

PR #604, opened 2018-04-16
Fix LPA writeback syntax

tomasonjo:
Moved direction parameter to config map for lpa writeback to be consistent with lpa.stream and all the rest

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/LabelPropagationProc.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/LabelPropagationProc.java
@@ -103,7 +103,7 @@ public final class LabelPropagationProc {
 
     @Procedure(value = "algo.labelPropagation.stream")
     @Description("CALL algo.labelPropagation.stream(label:String, relationship:String, " +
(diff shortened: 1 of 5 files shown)


jexp:
How do we handle backward compatibility here?

Esp. for things like blog posts / sandboxes / graphgists using the existing syntax ...

tomasonjo:
Does it make sense to have both options and then slowly deprecate the

jexp:
Yes.

mneedham:
@tomasonjo this one is still a WIP if I understood correctly?

tomasonjo:
Yes the issue is with backwards compatibility for all guides, tutorials, etc... 

we can't have two procedures with the same name, so I am kinda stuck, because I don't want to introduce a new name for labelPropagation writeback algorithm.

What I could do is to locate all guides and tutorials made by Neo and myself and then correct them once we would apply this PR... other then that not sure what could be the alternative

mneedham:
I put in a commit for this that handles backwards compatibility in the latest release

mneedham:
I put in a commit for this that handles backwards compatibility in the latest release

PR #613, opened 2018-04-26
Suppress/remove implementers section from documentation

davidoliverSP2:


davidoliverSP2:
@mneedham As discussed, I've stopped the implementers section from being built, in order to check that the docs build cleanly without it. Does this content need to be removed, or is it required in another location? (i.e. should it be removed **everywhere**, or just in the docs build?)

jexp:
We should definitely keep it for the online docs, but we can use `ifdef:` macros in adoc instead to disable them for the ebook/book.

mneedham:
@jexp David and I discussed that the majority of people using the graph algorithms docs are going to be users rather than implementers so I don't want to include the implementers section there because it's a lot of information that isn't relevant to them. We can put the implementers section on GitHub pages somewhere.

davidoliverSP2:
Closing, replaced by #638

PR #650, opened 2018-05-25
WIP: Sequential Relationship loading for Huge Graphs

### LINKED ISSUE #330: consider loading graphs by a sequential scan of the relationship store
especially when,

the store is not fully memory mapped
we're reading more than X% of all relationships (where X might be as low as 50) (get via count-store)
Reason: Reading by node and then relationship-chain is a random access into any page which might not be mapped and has to be loaded while another is evicted, which causes constant I/O pressure and reduces load performance to the speed of disk.

If we change this to loading the nodes and their degrees first in parallel (usually we have much fewer nodes, and more per page) to build up our structure / information.

Then we scan the relationshipstore exactly once, and check for the source or target-id to be in our id-map and put the other id in the appropriate slot in the adjacency list.

Nodes whose degrees have been reached can be sorted / compacted in the background (or save space via chunks of compressed vlongs).

knutwalker:
This implements a scan over the relationship store to load the graph data.

All our existing loaders all operate on a per-node scan which produces random access for the relationship store. The huge graph will now load by iterating over all relationships in the store order and accumulate them per node.
The main downside is that we don't know the degree in advance and wouldn't know when we've seen all relationships for a single note until we have loaded everything. This would use too much memory, so we do a first pass scan to load just the degrees and a second pass to load the actual relationships.

The scanner is single threaded to reduce the random access IO on the relationship store to a minimum and it will push relationships in batches (configurable based on the batchSize parameter) to a number of threads.
In order to prevent unlimited buffering of batches, which would essentially be the same as loading everything in memory, the number of batches that are allowed to be in flight is limited and the scanning thread might pause to allow the importer threads to catch up. This also required that all importer can actually run – If an importer is not started because it is waiting in the queue of the thread pool for a free thread, the whole system might deadlock: The scanner would wait for the importer to process its batches, that importer waits for the other importers to finish and those wait for the scanner to provide more data.
To prevent this, we adjust the concurrency parameter to make sure that all threads can actually be started at once and won't be queued (based on best effort).

Each importer thread can only that many nodes that would fit into single primitive arrays. The paging structure used for huge arrays until this PR with a fixed, static page size is changed so that every thread effectively produce the data for a single page. This helps to remove access to shared data amongst each importer thread and avoids a final copy step at the end, as we just wrap all single pages in the respective data structure.
Combined with the aforementioned limitation on thread size, we have a minimum requirement for the thread size to import a certain number of relationships, e.g. in order to load 30bn relationships, we need at least 15 or 16 threads available: https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/708d1fc406c37b80c5baaa11f69eb26a1a7e84f9/core/src/main/java/org/neo4j/graphalgo/core/huge/ThreadSizing.java#L27-L89

In order to reduce the memory usage during importing even further, we store the relationships compressed using a zig-zag-delta encoding. If we have all relationships for a node, we unpack and sort those and compress them again using variable-int-delta encoding, which is the same compression scheme that we have been using until now. I have tried various different approaches for this compression (e.g. GroupVarlong, based on what Google is (was) using for their index structure; {Fast}PFoR-{128,256} and straight bit-packing adapted for 64bit longs from the usual 32bit impls from in papers, all with and without additional delta encoding) and I have found that the {vint,zigzag}-delta encoding beats every other approach in terms of memory requirements, speed, and, most of all and by a large margin, simplicity. The main reason is that all the other compression algorithms are designed to be efficient for compressing data that is readily available but don't work so well if the data arrives in a streaming like fashion. I could improve on the final memory requirements with some of those compressions but at a cost of much more memory required _during_ the import and the final benefit was in the order of single-digit percents, often less than one percent. Since the other algos are tuned for {de,}compressing large amounts of data, the speed benefits would only be visible for super nodes and the overhead for smaller nodes that only have a couple of hundred relationships would mean that the typical use-case would actually have a worse performance. Therefore, I decided to keep the existing compression scheme, with the exception that I am now doing a block-wise decompression of var ints instead of reading every var int one-by-one. In the future, we could decide to have different encoding based on the node degree, much like Neo already does for dense nodes, though we would need a much larger threshold than 50.

I would like to add measurements and profiling data to this PR (over the next few days) before we can merge it.

Closes #330 .


jexp:
@knutwalker did you consider the following things:

- read the node store in batches to identify the pointers to the rel-store and rel-group store for each node
- put them in an array at a node-id position [0..n]
- sort a copy of the array so we have them in order, so we can basically scan forward using these id-offsets for reading chunks
- for rel-chains in the rel-group-store there is the degree stored as numeric value in the prev-pointer of the first record, so we can read that directly. 
- As the current threshold is 50 rels, I guess in many stores we have a high percentage of relationships in the rel-group store (we could even make an estimate based of the size/high-id of the rel-group store) to see if it's worth it.
- I'm still not sure the single reader is performant enough to saturate the IO channel, with multi-readers I got up to 2G/s reads
- and as we know the store structure (high-id) we can just chunk it up in batches and read each batch in parallel accumulating the relationships as we go per threads and then combining them at the end.

PR #655, opened 2018-06-04
[WIP] Louvain Phase 2 

mknblch:
The PR replaces all of our previous louvain algo implementations. 
- Phase 1 (previously called louvain) has been renamed to `org.neo4j.graphalgo.impl.louvain.ModularityOptimization`. It assigns a community id to each node as usual but the algorithm itself has been replaced because the old experimental implementation ran into locks. It now runs a modularity optimization on each thread until either the algo converges or its iteration threshold is reached. The process with the best modularity is taken for phase 2

- `org.neo4j.graphalgo.impl.louvain.Louvain` implements Phase 2 which takes the result of the modularity optimization and builds a new in-memory graph of it. Each cluster is represented as single node and all relationships between clusters are merged together by accumulating their weights. This usualy results in a smaller graph. The algo stores the nodes from the original graphand the cluster assigned to it for each round.

- after rebuilding the graph phase 1 is triggered again as long as the modularity keeps growing. if no further improvements can be made the algorithm terminates. the community structure with the best modularity is taken as final result.

the algo is nondeterministic and may lead to different results per ran.

**TODO**

- test on big graphs
- impl. streaming method


mneedham:
This is merged in by #684

PR #662, opened 2018-07-13
Added Ability to perform Random Walks

jexp:
* based on work from the students group at the HPI, see meta-exp.github.io
* supports random and node2vec strategy
* extended Graph API with getRelationship(node, index)
* `algo.randomWalk.stream(label, rel-type, start, steps, walks, {config}) yield nodes, path`
* Added basic documentation (not linked yet)

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/impl/walking/NodeWalker.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/impl/walking/NodeWalker.java
@@ -1,0 +1,246 @@
+package org.neo4j.graphalgo.impl.walking;
+
+import org.neo4j.graphalgo.api.Graph;
+import org.neo4j.graphalgo.api.Degrees;
+import org.neo4j.graphalgo.api.RelationshipConsumer;
+import org.neo4j.graphalgo.core.utils.ParallelUtil;
+import org.neo4j.graphalgo.core.utils.Pools;
+import org.neo4j.graphalgo.core.utils.QueueBasedSpliterator;
+import org.neo4j.graphalgo.core.utils.TerminationFlag;
+import org.neo4j.graphdb.*;
+import org.neo4j.logging.Log;
+import org.neo4j.procedure.Name;
+
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class NodeWalker {
+
+
+    public Stream<long[]> randomWalk(Graph graph, @Name(value = "steps", defaultValue = "80") int steps, NodeWalker.NextNodeStrategy strategy, TerminationFlag terminationFlag, int concurrency, int limit, PrimitiveIterator.OfInt idStream) {
+        int timeout = 100;
+        int queueSize = 1000;
+
+        int batchSize = ParallelUtil.adjustBatchSize(limit, concurrency, 100);
+        Collection<Runnable> tasks = new ArrayList<>((limit / batchSize) + 1);
+
+        ArrayBlockingQueue<long[]> queue = new ArrayBlockingQueue<>(queueSize);
+        long[] TOMB = new long[0];
+
+        while (idStream.hasNext()) {
+            int[] ids = new int[batchSize];
+            int i=0;
+            while (i<batchSize && idStream.hasNext()) {
+                ids[i++]=idStream.nextInt();
+            }
+            int size = i;
+            tasks.add(() -> {
+                for (int j = 0; j < size; j++) {
+                    put(queue,doWalk(ids[j], steps, strategy, graph, terminationFlag));
+                }
+            });
+        }
+        new Thread(() -> {
+            ParallelUtil.runWithConcurrency(concurrency, tasks, terminationFlag, Pools.DEFAULT);
+            put(queue,TOMB);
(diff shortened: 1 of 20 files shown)


algo/src/main/java/org/neo4j/graphalgo/impl/walking/WalkResult.java
```diff
@@ -0,0 +1,17 @@
+package org.neo4j.graphalgo.impl.walking;
+
+import org.neo4j.graphdb.Path;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class WalkResult {
+    public Path path;
+    public List<Long> nodes;
```
mneedham: Can we name this `nodeIds`? I can always do it after merging - just would make it more consistent with what we're doing elsewhere

algo/src/main/java/org/neo4j/graphalgo/walking/NodeWalkerProc.java
```diff
@@ -0,0 +1,167 @@
...
+
+        Boolean returnPath = configuration.get("path", false);
+
+        int limit = (walks == -1) ? nodeCount : Math.toIntExact(walks);
+
+        PrimitiveIterator.OfInt idStream = IntStream.range(0, limit).unordered().parallel().flatMap((s) -> idStream(start, graph, limit)).limit(limit).iterator();
+
+        Stream<long[]> randomWalks = new NodeWalker().randomWalk(graph, (int) steps, strategy, terminationFlag, concurrency, limit, idStream);
+        return randomWalks
+                .map( nodes -> new WalkResult(nodes, returnPath ? toPath(api, nodes,direction) : null));
```
mneedham: Should we return the nodeId where the random walk begins from as part of the stream as well, or should users just assume that is the first element of the `nodes` array?

algo/src/main/java/org/neo4j/graphalgo/walking/NodeWalkerProc.java
```diff
@@ -0,0 +1,167 @@
...
+    @Context
+    public Log log;
+
+    @Context
+    public KernelTransaction transaction;
+
+
+
+    @Procedure(name = "algo.randomWalk.stream", mode = Mode.READ)
+    @Description("CALL algo.randomWalk.stream(label, relationship, start:null=all/[ids]/label, steps, walks, {mode:random/node2vec, return:1.0, inOut:1.0, path:false/true concurrency:4, direction:'BOTH'}) " +
```
mneedham: I wonder if we should make this API the same as we do for the shortestpath procs where you pass in the label + relationship as config parameters. So it would then read:

```
public Stream<WalkResult> randomWalk(
             @Name(value = "start", defaultValue = "null") Object start,
             @Name(value = "steps", defaultValue = "10") long steps,
             @Name(value = "walks", defaultValue = "1") long walks,
             @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {

String label = config.get("nodeQuery");
String relationshipType = config.get("relationshipQuery")
....
```

I think that would be easier to understand as an API. We could perhaps also create a config parameter to go with the `start` parameter where you could indicate what you're intending to pass in as your starting point.

So it could default to 'label' but you could have other things too:

```
CALL algo.randomWalk.stream(node,....{startType: "node"})
CALL algo.randomWalk.stream(1,....{startType: "nodeId"})
```

Would save us having to do the type check on the value and we do a similar thing when allowing people to pass in label + reltype vs Cypher queries for graph projections.

algo/src/main/java/org/neo4j/graphalgo/impl/walking/NodeWalker.java
```diff
@@ -0,0 +1,246 @@
...
+    */
+
+    public static class RandomNextNodeStrategy extends NextNodeStrategy {
+
+        public RandomNextNodeStrategy(Graph graph, Degrees degrees) {
+            super(graph, degrees);
+        }
+
+        @Override
+        public int getNextNode(int currentNodeId, int previousNodeId) {
```
mneedham: If I'm reading correctly both of the provided strategies select one of a node's relationships to follow. Should we have a strategy where part of the random walk could involve jumping to a completely different node and continuing the walk from there?

algo/src/main/java/org/neo4j/graphalgo/walking/NodeWalkerProc.java
jexp: That was at least what the students used, and also in a path it's always the first node-id.
You definitely need to return it as most uses don't have a single start id.
But we can also add an extra column if that makes sense. I wouldn't want to remove it from the array though as that breaks the path.

algo/src/main/java/org/neo4j/graphalgo/impl/walking/WalkResult.java
jexp: Sure

algo/src/main/java/org/neo4j/graphalgo/impl/walking/NodeWalker.java
jexp: I was thinking about that, the main drawback I see is that it would kinda break the "path" in the graph but if we use virtual rels anyway that might be fine.
Not sure if we should somehow indicate in the results where such a jump occurred?

PR #664, opened 2018-07-16
store weights along with relationships for Yens k-shortest path

mneedham:
It makes it much easier to consume the output of Yens k-shortest path if we also store the distance on the generated relationships


mneedham:
@knutwalker I added a test to check that relationship weights get stored when traversing an undirected graph. It seems to work fine & I think this is the code that handles it:

https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/3.4/core/src/main/java/org/neo4j/graphalgo/core/heavyweight/VisitRelationship.java#L139

knutwalker:
@mneedham awesome, thank you. LGTM

PR #665, opened 2018-07-16
Cover edge-cases for nodes that are referencing themselves

knutwalker:
In the case where it was working, the additional relationship got swallowed by https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/c56b7748df88ae515ea9bfbbbd421573f315734d/core/src/main/java/org/neo4j/graphalgo/core/heavyweight/VisitRelationship.java#L76-L78 due to the different cursor traversal order, based on what relationships were created first. We saw a degree of 2 in both cased but got three relationships from the cursor, either [0, 1, 0] or  [1, 0, 0]. it worked by accident for the second case.


mneedham:
LGTM

PR #669, opened 2018-07-18
WIP: Personalized PageRank

mneedham:


algo/src/main/java/org/neo4j/graphalgo/PageRankProc.java
```diff
@@ -177,10 +186,14 @@ private PageRankResult evaluate(
         final int concurrency = configuration.getConcurrency(Pools.getNoThreadsInDefaultPool());
         log.debug("Computing page rank with damping of " + dampingFactor + " and " + iterations + " iterations.");
 
+
+        List<Node> sourceNodes = configuration.get("sourceNodes", new ArrayList<>());
+        Stream<Long> sourceNodeIds = sourceNodes.stream().map(Node::getId);
```
knutwalker: We could use `mapToLong` to avoid boxing to Long instances

algo/src/main/java/org/neo4j/graphalgo/impl/HugePageRank.java
```diff
@@ -209,6 +205,7 @@ private void initializeSteps() {
                 concurrency,
                 idMapping.nodeCount(),
                 dampingFactor,
+                sourceNodeIds.map(graph::toMappedNodeId).collect(Collectors.toList()),
```
knutwalker: Since this is `Huge`, we should use `toHugeMappedNodeId` to get the internal long id.

knutwalker: If we use the LongStream here, `toArray` would probably better.

knutwalker: Also, a Node may not have been mapped if there was any Label/Type filtering, so maybe add a `filter(mappedId -> mappedId != -1L)` to remove the nodes that we don't know about.

algo/src/main/java/org/neo4j/graphalgo/PageRankProc.java
mneedham: Have updated this now

algo/src/main/java/org/neo4j/graphalgo/impl/HugePageRank.java
```diff
@@ -606,8 +603,21 @@ private void initialize() {
...
+                Arrays.fill(partitionRank, alpha);
+            } else {
+                Arrays.fill(partitionRank,0);
+
+                long[] partitionSourceNodeIds = LongStream.of(sourceNodeIds)
+                        .filter(sourceNodeId -> sourceNodeId >= startNode && sourceNodeId <= endNode)
+                        .toArray();
+
+                for (long sourceNodeId : partitionSourceNodeIds) {
+                    partitionRank[Math.toIntExact(sourceNodeId - this.startNode)] = alpha;
```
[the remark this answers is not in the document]
knutwalker: @mneedham yes, that's fine; we usually have `(int) (nodeId - startNode)` for local nodes or `(int) (nodeId - starts[partitionIndex])` for nodes in a different partition.

knutwalker:
LGTM

PR #687, opened 2018-08-31
write version of jaccard

mneedham:


algo/src/main/java/org/neo4j/graphalgo/JaccardProc.java
```diff
@@ -33,24 +54,161 @@
...
+        TerminationFlag terminationFlag = TerminationFlag.wrap(transaction);
+
+        int concurrency = configuration.getConcurrency();
+
+        return jaccardStreamMe(ids, length, terminationFlag, concurrency, similarityCutoff);
+    }
+
+    @Procedure(name = "algo.jaccard", mode = Mode.WRITE)
+    @Description("CALL algo.jaccard([{source:id, targets:[ids]}], {similarityCutoff:-1,degreeCutoff:0}) " +
+            "YIELD percentile50, percentile75, percentile90, percentile99, percentile999, percentile100 - computes jaccard similarities")
```
jexp: can we shorten these to pXX

algo/src/main/java/org/neo4j/graphalgo/JaccardProc.java
```diff
@@ -33,24 +54,161 @@
...
+        Stream<SimilarityResult> stream = jaccardStreamMe(ids, (int) length, terminationFlag, concurrency, similarityCutoff);
+
+        AtomicLong similarityPairs = new AtomicLong();
+
+        if(configuration.isWriteFlag() && similarityCutoff > 0.0) {
+            SimilarityExporter similarityExporter = new SimilarityExporter(api,
+                    configuration.get("relationshipType", "SIMILAR"),
+                    configuration.getWriteProperty("score"));
+
+            Stream<SimilarityResult> similarities = stream.peek(result -> {
```
jexp: we can probably wrap the lambda into a local function or variable.

algo/src/main/java/org/neo4j/graphalgo/JaccardProc.java
jexp: p95 would be good to have.

mneedham: Oh yeh that's much better. T'is done

algo/src/main/java/org/neo4j/graphalgo/JaccardProc.java
mneedham: Yup. Done.

PR #695, opened 2018-09-07
include all procs regardless of package (random walk was one level down)

mneedham:


mneedham:
@davidoliverSP2 this should fix the issue with the random walk procedure not showing on the list of procedures page

PR #696, opened 2018-09-09
one hot encoding function to simplify getting categorical data into r…

mneedham:
…equired format

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/ml/OneHotEncoding.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/ml/OneHotEncoding.java
@@ -4,19 +4,30 @@ import org.neo4j.procedure.Description;
 import org.neo4j.procedure.Name;
 import org.neo4j.procedure.UserFunction;
 
+import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


algo/src/main/java/org/neo4j/graphalgo/ml/OneHotEncoding.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.stream.LongStream;
+
+public class OneHotEncoding {
+
+    @UserFunction("algo.ml.oneHotEncoding")
+    @Description("CALL algo.ml.oneHotEncoding(availableValues, selectedValues) - return a list of selected values in a one hot encoding format.")
+    public List<Long> oneHotEncoding(@Name(value = "availableValues") List<Object> availableValues,
+                                     @Name(value = "selectedValues") List<Object> selectedValues) {
+        if (availableValues == null) {
+            return LongStream.empty().boxed().collect(Collectors.toList());
```
jexp: return Collections.emtpyList() would have been enough :)

PR #707, opened 2018-09-17
handle arrays of values - this is what we get if we store properties

mneedham:
@jexp not sure if there's an easier way than this?

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityProc.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityProc.java
@@ -214,7 +214,22 @@ public class SimilarityProc {
         WeightedInput[] inputs = new WeightedInput[data.size()];
         int idx = 0;
         for (Map<String, Object> row : data) {
-            List<Number> weightList = (List<Number>) row.get("weights");
+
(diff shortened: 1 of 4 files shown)


algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityProc.java
```diff
@@ -214,7 +214,22 @@ Double getSimilarityCutoff(ProcedureConfiguration configuration) {
         WeightedInput[] inputs = new WeightedInput[data.size()];
         int idx = 0;
         for (Map<String, Object> row : data) {
-            List<Number> weightList = (List<Number>) row.get("weights");
+
```
jexp: I think this is ok, should also be fastest. We should probably pull this out into a util function.

An alternative (but slower) would be to check if the class of the type is an Array and the component-type is a subtype is number and iterate with reflection.

jexp: We will also need it for Euclidean.

mneedham: Added a test for Euclidean and updated our Jaccard processing code too

PR #709, opened 2018-09-18
Direct id map

jexp:


mneedham:
I think this one needs to be rebased as the cosine/euclidean stuff is merged into 3.4 now

PR #710, opened 2018-09-18
shortest paths now undirected

mneedham:


doc/asciidoc/algorithms-path-finding.adoc
```diff
@@ -20,6 +20,15 @@ The following path finding algorithms help find the shortest path or evaluate th
 * <<algorithms-a_star, A*>> (`algo.shortestPath.astar`)
 * <<algorithms-yens-k-shortest-path, Yen’s K-shortest paths>> (`algo.kShortestPaths`)
 
+[WARNING]
+====
+As of the 3.4.7.0 release, all the Shortest Path procedures assume that they are being executed on undirected graphs
```
davidoliverSP2: Suggested reword:

> As of the 3.4.7.0 release, all the Shortest Path procedures assume that they are being executed on undirected graphs. Therefore, the direction of a relationship can be ignored.

doc/asciidoc/algorithms-path-finding.adoc
```diff
@@ -20,6 +20,15 @@ The following path finding algorithms help find the shortest path or evaluate th
...
+i.e. the direction of a relationship can be ignored
+
+If you are running these algorithms on a graph where the direction is important you can override this by passing in the `direction` parameter
```
davidoliverSP2: Suggested rewording:

> If you are running these algorithms on a graph where the direction is important, you can use the `direction` parameter. For example, `direction:"INCOMING"` or `direction:"OUTGOING"`.

PR #711, opened 2018-09-19
Ensure that documentation includes new categories

davidoliverSP2:
The new categories (Similarity and Preprocessing) were inconsistent in the order of presentation, and needed to be added to the content map.


mneedham:
LGTM

PR #712, opened 2018-09-20
Fix yens-ksp queue usage

mknblch:


tests/src/test/java/org/neo4j/graphalgo/impl/KShortestPathsUserTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * @author mknblch
+ *
+ * TODO rm if verified
+ */
+@Ignore
+public class KShortestPathsUserTest {
+
+    private static final String PATH = "/Users/mknobloch/Downloads";
```
kcrimson: hardcoded path to directory, are you going to remove this per comment as I understand?

mneedham: @kcrimson yeh that's only been left there temporarily as this is WIP and I've been testing it with your dataset before we merge the PR in. We won't merge it in with that hardcoded path or the build won't work :)

mknblch:
see #725 for more info

PR #715, opened 2018-09-23
Balanced triads docs

tomasonjo:


mneedham:
I think all this is fine, but the way we've implemented it right now doesn't match up with how it'll be documented. e.g.

> Unlike other algorithms where there are only positive relationships available, balance theory differentiates between positive and negative relationships.

But right now we work out the positivity of a relationship based on a weight property on a single relationship type. I'm not sure what the best way to do that is.

Should it be:

```
CALL algo.balancedTriads('Person', null, {weightProperty:'weight', positiveRelationship: "FRIEND", negativeRelationship: "ENEMY"}) 
```

and then if you don't specify the positive/negative types then we read the weight property?

```
CALL algo.balancedTriads('Person', 'TYPE', {weightProperty:'weight'}) 
```

or maybe:

```
CALL algo.balancedTriads('Person', "FRIEND", "ENEMY") 
```

which would mean if you used the same relationship type for both then it would refer to the weight property:

```
CALL algo.balancedTriads('Person', "TYPE", "TYPE") 
```

I think I like the first approach, which would mean we can leave the current logic how it is, but if you specify `positiveRelationship` and `negativeRelationship` you should set the 2nd parameter as `null` and then we're gonna ignore the weight property and instead assign 1.0 to positive and -1.0 to negative.

What do you think?

Also cc @jexp @mknblch

PR #719, opened 2018-09-27
3.4 degree centrality

mneedham:


mneedham:
Gonna merge this so that we can use it with weighted PR. Not hooking it up to procedures until we figure out the best way to handle the different ways that our graph loaders treat multiple relationships between pairs of nodes

Issue #729, opened 2018-10-02
Jaccard index store only topK doesnt work


tomasonjo:
Create data:
```
MERGE (french:Cuisine {name:'French'})
MERGE (italian:Cuisine {name:'Italian'})
MERGE (indian:Cuisine {name:'Indian'})
MERGE (lebanese:Cuisine {name:'Lebanese'})
MERGE (portuguese:Cuisine {name:'Portuguese'})

MERGE (zhen:Person {name: "Zhen"})
MERGE (praveena:Person {name: "Praveena"})
MERGE (michael:Person {name: "Michael"})
MERGE (arya:Person {name: "Arya"})
MERGE (karin:Person {name: "Karin"})

MERGE (praveena)-[:LIKES]->(indian)
MERGE (praveena)-[:LIKES]->(portuguese)

MERGE (zhen)-[:LIKES]->(french)
MERGE (zhen)-[:LIKES]->(indian)

MERGE (michael)-[:LIKES]->(french)
MERGE (michael)-[:LIKES]->(italian)
MERGE (michael)-[:LIKES]->(indian)

MERGE (arya)-[:LIKES]->(lebanese)
MERGE (arya)-[:LIKES]->(italian)
MERGE (arya)-[:LIKES]->(portuguese)

MERGE (karin)-[:LIKES]->(lebanese)
MERGE (karin)-[:LIKES]->(italian)
```

If I want to set the only parameter when storing relationships is topK, then it doesn't actually store results, though the results seem fine
```
MATCH (p:Person)-[:LIKES]->(cuisine)
WITH {item:id(p), categories: collect(id(cuisine))} as userData
WITH collect(userData) as data
CALL algo.similarity.jaccard(data, {topK: 1, write:true})
YIELD nodes, similarityPairs, write, writeRelationshipType, writeProperty, min, max, mean, stdDev, p25, p50, p75, p90, p95, p99, p999, p100
RETURN nodes, similarityPairs, write, writeRelationshipType, writeProperty, min, max, mean, p95
```

If I add similarityCutoff then it works
```

MATCH (p:Person)-[:LIKES]->(cuisine)
WITH {item:id(p), categories: collect(id(cuisine))} as userData
WITH collect(userData) as data
CALL algo.similarity.jaccard(data, {topK: 1, similarityCutoff: 0.1, write:true})
[...]

PR #730, opened 2018-10-02
Fix the write boolean for similarity index

tomasonjo:
Not sure if there is a deeper reason for `similarityCutoff` to have to be greater than 0.0 to allow writeback. Fixes https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/729


mneedham:
The extra condition is deliberate as we thought it doesn't make sense to write a `SIMILAR` relationship between two nodes if they aren't actually similar in any way. 

I think we should change the default to be -1 for all of them, but then update the documentation so that it's clearer what you have to do to write relationships.

I'm not really opposed to what you're suggesting either though so I guess @jexp can be the tie breaker!

jexp:
I'm not sure 100% sure what you mean.

We should never write relationships for dissimilar pairs as it just explodes with n^2.
I want the defaults to be safe, i.e. if you don't specify something it should not hurt you.

The rule is different depending on the type of similarity.
So we probably only have to fix the default for cosine or make it by `similarity != default`.

tomasonjo:
What I would suggest then is the following:

We got four similarities with possible values:
- cosine: -1 to 1
- jaccard: 0 to 1
- overlap: 0 to 1
- euclidian: 0 to ∞

If we keep it simple and just use one default `similarityCutoff` I would use 0.5 as cosine can still blow up with 0.0 still and we cover jaccard and overlap with it. 

On the other hand it is very hard to set an arbitrate default number for euclidian, so maybe we just stick with 0.5.

We can still include the boolean condition for cosine writeback to be greater than 0.0 and never allow user to writeback dissimilar pairs, but need to include it in the documentation then. 

WDYT?

tomasonjo:
Much cleaner option might be setting default `topK` parameter at 20 and not set any default `similarityCutoff`. This handles the default version very nicely IMO

tomasonjo:
I changed the default `similarityCutoff` to 0.1. This implies to all except for euclidian distance where I left the default value -1.0.

mneedham:
I think what we'll do is throw an error if you try and set `write: true` with a similarity of <= 0.0

Issue #736, opened 2018-10-09
Update Louvain docs


tomasonjo:
With the level 2 louvain we now store community of node at each iteration and hence the result is no more an integer, but an array of integers. Also param community has been changed to communities for streaming version

jexp:
Yes I fully agree with Tomaz, the single value param should stay (also for indexing)
And the array (which is more expensive) should be an optional, additional property.

tomasonjo:
Will the last two community ids in the communities array always be the same as the last pass always determines that the algo converged as the values haven't changed and stops iterations.

jexp:
@mknblch we MUST NOT change signatures of procedures (parameters or result columns) that are not backwards compatible.

So keep "community" for both the write and stream version and make the new communities an optional result via config (can default to true, but people need to be aware that much more data is written and streamed back).

See: https://community.neo4j.com/t/graph-algorithms/2333

mneedham:
@jexp  I think I made that change rather than @mknblch. We can make the change as you suggested, but the meaning of 'community' will be a bit different than it was before we implemented 2 stage Louvain.

In the 1 stage Louvain 'community' was referring to the communities found at stage 1 of the algorithm, whereas if we make it use the value from the last iteration of the algorithm that will be different, assuming that it runs for > 1 stage. 

Is it still ok to store the last stage as 'community' given that?

tomasonjo:
closed with https://github.com/neo4j-contrib/neo4j-graph-algorithms/pull/737

PR #737, opened 2018-10-10
3.4 louvain backward

mneedham:
Fixing backwards compatibility things as described in https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/736

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/impl/louvain/Louvain.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/impl/louvain/Louvain.java
@@ -209,15 +209,18 @@ public class Louvain extends Algorithm<Louvain> {
                 .mapToObj(i -> new Result(i, communities[i]));
     }
 
-    public Stream<StreamingResult> dendrogramStream() {
+    public Stream<StreamingResult> dendrogramStream(boolean includeIntermediateCommunities) {
         return IntStream.range(0, rootNodeCount)
                 .mapToObj(i -> {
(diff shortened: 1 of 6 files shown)


algo/src/main/java/org/neo4j/graphalgo/impl/louvain/Louvain.java
```diff
@@ -209,15 +209,18 @@ public long getCommunityCount() {
...
 
-    public Stream<StreamingResult> dendrogramStream() {
+    public Stream<StreamingResult> dendrogramStream(boolean includeIntermediateCommunities) {
         return IntStream.range(0, rootNodeCount)
                 .mapToObj(i -> {
-                    List<Long> communities = new ArrayList<>(dendrogram.length);
-                    for (int[] community : dendrogram) {
-                        communities.add((long) community[i]);
+                    List<Long> communitiesList = null;
+                    if (includeIntermediateCommunities) {
```
tomasonjo: Maybe just a cosmetic fix, but won't the community ids of the last two iterations always be the same as the algo checks if it has converged and if the community ids don't change in the last iteration it stops. So we can skip returning the last value of array as it will be always the same as second last value.

mneedham: @mknblch will know for sure so Martin what do you say?

tomasonjo: tested it on another dataset and it looks like it is not always the same, so please just disregard this

PR #747, opened 2018-10-25
3.4 feature vectors

mneedham:


mneedham:
The problem we were trying to solve was done using Run Length Encoding on the similarity algorithms instead. Maybe we can revisit this one later

PR #756, opened 2018-11-13
[WIP] Louvain predefined communities

mknblch:


mneedham:
There are a couple of things that we still need to address:

1. If you run the algorithm again while specifying pre-defined communities on the same graph it will throw an NPE. This is the offending line - https://github.com/neo4j-contrib/neo4j-graph-algorithms/blob/3.4/algo/src/main/java/org/neo4j/graphalgo/LouvainProc.java#L151 We can fix that by reading from `finalCommunities` instead. 

2. Any nodes that don't have predefined communities get put into the same community (`-1`). This will bias their end community, so they should all start in a unique community (some value greater than the largest pre defined community seems like the easiest way to do that)

PR #758, opened 2018-11-16
Add surefire plugin to pom.xml

tomasonjo:


mneedham:
@tomasonjo  I forgot why we need this plugin added to the POM - what thing does it fix?

tomasonjo:
At the time I couldn't build the algos locally without this plugin. Tried it now on both 3.4 and 3.5 and it works fine so I'll just close this

PR #765, opened 2018-12-03
Fix library header for 3.5

mariascharin:


mariascharin:
@mneedham This fixes some incorrect behavior of the Docs library header.

PR #769, opened 2018-12-08
Pearson

mneedham:


mneedham:
These are the benchmark results for Pearson vs Cosine:

```
Benchmark                                     Mode   Cnt        Score        Error  Units
PearsonSimilarityBenchmark.cosineSquares        ss  1000  1749163.162 ±  65018.859  ns/op
PearsonSimilarityBenchmark.pearsonSquares       ss  1000  3408866.198 ±  88648.485  ns/op
PearsonSimilarityBenchmark.pearsonSquaresRle    ss  1000  6484999.724 ± 225489.707  ns/op
```

algo/src/main/java/org/neo4j/graphalgo/similarity/Similarities.java
```diff
@@ -67,6 +67,35 @@ public double cosineSimilarity(@Name("vector1") List<Number> vector1, @Name("vec
...
+            throw new RuntimeException("Vectors must be non-empty and of the same size");
+        }
+
+        double vector1Mean = vector1.stream().mapToDouble(Number::doubleValue).average().orElse(1);
+        double vector2Mean = vector2.stream().mapToDouble(Number::doubleValue).average().orElse(1);
+
+        double dotProductMinusMean = 0d;
+        double xLength = 0d;
+        double yLength = 0d;
+        for (int i = 0; i < vector1.size(); i++) {
```
jexp: we should change this to a `while` + 2 list iterators.

PR #779, opened 2019-01-07
Adamic Adar and Resource Allocation Link Prediction Algorithms

mneedham:


mneedham:
@davidoliverSP2 this is a new category for the algorithms. These PR has the code for them as well as the docs - hence why so many files changed!

readme.adoc
```diff
@@ -73,6 +73,12 @@ These algorithms help calculate the similarity of nodes:
 * <<algorithms-similarity-euclidean, Euclidean Distance>> (`algo.similarity.euclidean`)
 * <<algorithms-similarity-overlap, Overlap Similarity>> (`algo.similarity.overlap`)
 
+=== Link Prediction
+
+These algorithms can be used as part of Link Prediction solution:
+
+* <<algorithms-linkprediction-adamic-adar, Adamic Adar Similarity>> (`algo.similarity.adamicAdar`)
+
```
davidoliverSP2: should the **Resource Allocation algorithm** feature here too?

doc/asciidoc/linkprediction-adamic-adar.adoc
```diff
@@ -0,0 +1,105 @@
...
+
+The library contains a function to calculate similarity between two nodes.
+
+// end::explanation[]
+
+[[algorithms-linkprediction-adamic-adar-usecase]]
+== Use-cases - when to use the Adamic Adar Similarity algorithm
+
+// tag::use-case[]
+We can use the Adamic Adar Similarity algorithm to work out the similarity between two nodes.
```
davidoliverSP2: We need to be careful with the consistency of naming this algorithm. It appears to be titled **The Adamic Adar algorithm**, but here it is titled **the Adamic Adar Similarity algorithm**. Above, it mentions _Adamic Adar similarity_, with a lower case "s".

I would recommend that we remove the "Similarity" from this example, and elsewhere in the page refer to only **The Adamic Adar algorithm** when we refer to "the algorithm".

doc/asciidoc/linkprediction-resource-allocation.adoc
```diff
@@ -0,0 +1,105 @@
...
+where `N(u)` is the set of nodes adjacent to `u`.
+
+A value of 0 indicates that two nodes are dissimilar, while higher values indicate nodes are more similar.
+
+The library contains a function to calculate similarity between two nodes.
+
+// end::explanation[]
+
+[[algorithms-linkprediction-resource-allocation-usecase]]
+== Use-cases - when to use the Resource Allocation Similarity algorithm
```
davidoliverSP2: This page has the same issue regarding consistency of name.

Here, we use **the Resource Allocation Similarity algorithm**, whereas the title is **the Resource Allocation algorithm**. I would suggest sticking to the titled version, unless there is a good reason not to, but consistency is best practice.

readme.adoc
mneedham: Yes! Also I have the name of the procedure wrong so I've updated that as well

doc/asciidoc/linkprediction-resource-allocation.adoc
mneedham: I have updated this to remove 'Similarity' everywhere. I was originally going to include this algorithms in the 'similarity' category, but realised that the way they're used is completely different and it didn't make sense to have them there.

doc/asciidoc/linkprediction-adamic-adar.adoc
mneedham: Have removed similarity for these as well

algo/src/main/java/org/neo4j/graphalgo/linkprediction/LinkPrediction.java
```diff
@@ -0,0 +1,136 @@
...
+        return sumInverse(vector);
+    }
+
+
+    public double sumInverse(@Name("vector") List<Number> vector) {
+        // https://arxiv.org/pdf/0901.0553.pdf
+        if (vector == null) return 0;
+
+        double score = 0.0;
+        for (int i = 0; i < vector.size(); i++) {
```
jexp: foreach

algo/src/main/java/org/neo4j/graphalgo/linkprediction/LinkPrediction.java
```diff
@@ -0,0 +1,136 @@
...
+                                       @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
+        ProcedureConfiguration configuration = ProcedureConfiguration.create(config);
+        RelationshipType relationshipType = configuration.getRelationship();
+        Direction direction = configuration.getDirection(Direction.BOTH);
+
+        List<Number> vector = new ArrayList<>();
+        try(Transaction tx = api.beginTx()) {
+            Set<Node> neighbors = findPotentialNeighbors(node1, relationshipType, direction);
+            neighbors.removeIf(node -> noCommonNeighbors(node, relationshipType, direction, node2));
+            neighbors.forEach(neighbor -> vector.add(degree(relationshipType, direction, neighbor)));
```
jexp: mapToDoulbe( nb -> 1.0/ degre(nb) ).sum()

algo/src/main/java/org/neo4j/graphalgo/linkprediction/LinkPrediction.java
```diff
@@ -0,0 +1,136 @@
...
+    @Description("algo.linkprediction.resourceAllocation(node1:Node, node2:Node, {relationshipQuery:'relationshipName', direction:'BOTH'}) " +
+            "given two nodes, calculate Resource Allocation similarity")
+    public double resourceAllocationSimilarity(@Name("node1") Node node1, @Name("node2") Node node2,
+                                       @Name(value = "config", defaultValue = "{}") Map<String, Object> config) {
+        ProcedureConfiguration configuration = ProcedureConfiguration.create(config);
+        RelationshipType relationshipType = configuration.getRelationship();
+        Direction direction = configuration.getDirection(Direction.BOTH);
+
+        List<Number> vector = new ArrayList<>();
+        try(Transaction tx = api.beginTx()) {
```
jexp: why a new tx?

algo/src/main/java/org/neo4j/graphalgo/linkprediction/LinkPrediction.java
```diff
@@ -0,0 +1,136 @@
...
+import org.neo4j.procedure.UserFunction;
+
+import java.util.*;
+
+public class LinkPrediction {
+
+    @Context
+    public GraphDatabaseAPI api;
+
+    @UserFunction("algo.linkprediction.adamicAdar")
```
jexp: do you plan this also as a proc on the algo-graph?
then we could use intersection etc.

mneedham: Wasn't planning to as this one and resource allocation are used to find closeness between pairs of nodes. I don't think you'd use them to do all pairs similarity like we do with the similarity algos (Pearson, Cosine, Jaccard, etc)

algo/src/main/java/org/neo4j/graphalgo/linkprediction/LinkPrediction.java
mneedham: I always forget that functions have an implicit tx!

algo/src/main/java/org/neo4j/graphalgo/linkprediction/LinkPrediction.java
mneedham: Nice. Have updated

PR #798, opened 2019-01-25
unify community algo results

mknblch:


mneedham:
This is already merged in via another commit.

PR #808, opened 2019-02-05
3.4 batching similarity writes

mneedham:


mneedham:
@jexp  this is the simplest approach we can take. It doesn't do anything clever with retries. Since we're doing it all single threaded we won't have deadlock issues to take care of, so I think it should be ok

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityExporter.java
```diff
@@ -62,6 +62,29 @@ private void export(SimilarityResult similarityResult) {
...
+    private void createRelationship(SimilarityResult similarityResult, KernelTransaction statement) throws EntityNotFoundException, InvalidTransactionTypeKernelException, AutoIndexingKernelException {
+        long node1 = similarityResult.item1;
+        long node2 = similarityResult.item2;
+        long relationshipId = statement.dataWrite().relationshipCreate(node1, relationshipTypeId, node2);
+
+        statement.dataWrite().relationshipSetProperty(
+                relationshipId, propertyId, Values.doubleValue(similarityResult.similarity));
+    }
+
     private int getOrCreateRelationshipId(String relationshipType) {
```
jexp: getOrCreateRelationshipId -> getOrCreateRelationshipTypeId

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityExporter.java
```diff
@@ -62,6 +62,29 @@ private void export(SimilarityResult similarityResult) {
 
     }
 
+    private void export(List<SimilarityResult> similarityResults) {
+        applyInTransaction(statement -> {
+            for (SimilarityResult similarityResult : similarityResults) {
+                try {
+                    createRelationship(similarityResult, statement);
+                } catch (KernelException e) {
+                    ExceptionUtil.throwKernelException(e);
```
jexp: should we really abort the whole tx if one relationship failed?

What do you think? We could also have a counter and we only throw after a certain percentage failed?
and otherwise log them.
In algos I think we don't abort writes if a few fail b/c those are all independent tx.

Not sure though if after a single KernelException the TX is tainted/failed anyway.
Perhaps something to turn into an issue and handle later.
Also that Exception int the callback where does it land?
In `applyInTransaction` and is it logged their or thrown up?

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityExporter.java
```diff
@@ -74,8 +97,23 @@ private int getOrCreatePropertyId(String propertyName) {
                 .propertyKeyGetOrCreateForName(propertyName));
     }
 
-    private void writeSequential(Stream<SimilarityResult> similarityPairs) {
-        similarityPairs.forEach(this::export);
+    private void writeSequential(Stream<SimilarityResult> similarityPairs, long batchSize) {
```
jexp: Would be really amazing if we could parallelize that.

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityProc.java
```diff
@@ -75,7 +74,12 @@ Long getDegreeCutoff(ProcedureConfiguration configuration) {
         return configuration.get("degreeCutoff", 0L);
     }
 
-    Stream<SimilaritySummaryResult> writeAndAggregateResults(Stream<SimilarityResult> stream, int length, boolean write, String writeRelationshipType, String writeProperty) {
+    Long getWriteBatchSize(ProcedureConfiguration configuration) {
+        return configuration.get("writeBatchSize", 1000L);
```
jexp: 1000 seems on the very low end esp. if we write sequentially, I'd up it to at least 10k or 100k

doc/asciidoc/similarity-cosine.adoc
```diff
@@ -284,6 +284,7 @@ YIELD nodes, similarityPairs, write, writeRelationshipType, writeProperty, min,
 | `concurrency`            | int     | available CPUs | yes      | The number of concurrent threads.
 | `graph`                  | string  | dense          | yes      | The graph name ('dense' or 'cypher').
 | `write`                  | boolean | false          | yes      | Indicates whether results should be stored.
+| `writeBatchSize`            | int     | 1000 | yes      | The batch size to use when storing results.
```
jexp: update default here too

doc/asciidoc/similarity-euclidean.adoc
```diff
@@ -273,6 +273,7 @@ YIELD nodes, similarityPairs, write, writeRelationshipType, writeProperty, min,
 | `concurrency`            | int     | available CPUs | yes      | The number of concurrent threads.
 | `graph`                  | string  | dense          | yes      | The graph name ('dense' or 'cypher').
 | `write`                  | boolean | false          | yes      | Indicates whether results should be stored.
+| `writeBatchSize`            | int     | 1000 | yes      | The batch size to use when storing results.
```
jexp: here too

doc/asciidoc/similarity-jaccard.adoc
```diff
@@ -256,6 +256,7 @@ include::scripts/similarity-jaccard.cypher[tag=query]
 | `degreeCutoff`           | int     | 0              | yes      | The threshold for the number of items in the `targets` list. If the list contains less than this amount, that node will be excluded from the calculation.
 | `concurrency`            | int     | available CPUs | yes      | The number of concurrent threads.
 | `write`                  | boolean | false          | yes      | Indicates whether results should be stored.
+| `writeBatchSize`            | int     | 1000 | yes      | The batch size to use when storing results.
```
jexp: here too

doc/asciidoc/similarity-pearson.adoc
```diff
@@ -295,6 +295,7 @@ YIELD nodes, similarityPairs, write, writeRelationshipType, writeProperty, min,
 | `concurrency`            | int     | available CPUs | yes      | The number of concurrent threads.
 | `graph`                  | string  | dense          | yes      | The graph name ('dense' or 'cypher').
 | `write`                  | boolean | false          | yes      | Indicates whether results should be stored.
+| `writeBatchSize`            | int     | 1000 | yes      | The batch size to use when storing results.
```
jexp: and here :)

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityExporter.java
mneedham: Yeh...I dunno if there's a way that we can partition the updates to run them in parallel without having any deadlocks.I suppose we need to create batches based on running connected components over the similarity graph that's going to be created?!

jexp: Yep that's what I thought. Might be easier if we had the graph loaded to run the algo. Or we could load it temporarily to compute clusters then write the "inner" cluster rels in parallel first and then the "border" rels in a single batch.

Would be a fun experiment.

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityExporter.java
mneedham: After we looked at the code this seems consistent with what is done elsewhere in the library, but we might look at changing that all in another commit

algo/src/main/java/org/neo4j/graphalgo/similarity/SimilarityExporter.java
mneedham: We have this on another branch that's still WIP. I think we can merge this sequential batching and add the parallel writing afterwards

doc/asciidoc/similarity-jaccard.adoc
mneedham: Fixed all thse

PR #814, opened 2019-02-11
batching wip

mneedham:


mneedham:
Testing on the recommendations dataset, there are so many partitions that the sequential approach is still faster.

selection_249

jexp:
Hmm do we want to add this as an option to be able to experiment more with it?

mneedham:
We can merge this in but I need to fix the conflicts first

PR #818, opened 2019-02-14
put louvain concurrency to 1 until we have a parallel implementation

mneedham:


mneedham:
@jexp I think this is the idea right? Algorithm on a single thread, but all other use of concurrency respect whatever config you pass in

PR #819, opened 2019-02-18
Algo rnlouvain3.4

mknblch:


mneedham:
@mknblch Rather then the config being `randomNeighbor:false`, can we instead make it more generic. 

So like this instead:

`communitySelection: "classic"` - default, uses the standard Louvain approach 
`communitySelection: "random"` - random neighbors approach

PR #821, opened 2019-02-19
Dave fauth docs updates

mneedham:


mneedham:
@davidoliverSP2 This PR addresses both the issues that Dave raised. While doing it I realised that we had the same text about Cypher projections pasted everywhere so I pulled it out into one place and referenced it.

mneedham:
sweet - thanks! And yeh all looks good to me

mneedham:
@davidoliverSP2 I've merged that into 3.5 as well so we can publish away whenever is good

PR #829, opened 2019-02-26
Bug/k shortest path max depth

frant-hartm:


mneedham:
Thanks! Will merge this one in

PR #830, opened 2019-02-28
[WIP] Degree Centrality

mneedham:


mneedham:
@jexp  we had two tests that weren't using the new duplicate strategy. Have updated both of those, so we should be good to merge this now

PR #834, opened 2019-03-04
eigenvector centrality

mneedham:


mneedham:
@jexp this now has support for the different normalisations

mneedham:
[ ] Implement the different normalisations for partitioned results

algo/src/main/java/org/neo4j/graphalgo/CentralityUtils.java
```diff
@@ -0,0 +1,66 @@
...
+                        .build();
+                result.export(propertyName, exporter);
+            }
+            statsBuilder.withWrite(true).withProperty(propertyName);
+        } else {
+            statsBuilder.withWrite(false);
+        }
+    }
+
+    static Stream<CentralityScore> streamResults(Graph graph, CentralityResult scores) {
```
jexp: could we add a parallel(Stream) here?

algo/src/main/java/org/neo4j/graphalgo/impl/results/NormalizedCentralityResult.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public double computeMax() {
+        return result.computeMax();
+    }
+
+    @Override
+    public double computeL2Norm() {
+        return 0;
```
jexp: can this lead to division by zero somewhere?

tests/src/test/java/org/neo4j/graphalgo/algo/EigenvectorCentralityProcIntegrationTest.java
```diff
@@ -0,0 +1,299 @@
...
+                        "ORDER BY score DESC " +
+                        "LIMIT 10",
+                row -> actual.put(
+                        (Long)row.get("nodeId"),
+                        (Double) row.get("score")));
+
+        assertMapEquals(expected, actual);
+    }
+
+//    @Test
```
jexp: can we delete these or do you want to keep them for weighted EV ?

algo/src/main/java/org/neo4j/graphalgo/CentralityUtils.java
mneedham: I suppose so - can't think why not. This code has already been there for ages - I've just moved it out into one place

tests/src/test/java/org/neo4j/graphalgo/algo/EigenvectorCentralityProcIntegrationTest.java
mneedham: yeh I was keeping them so that when I add weighted EV I'll uncomment them

algo/src/main/java/org/neo4j/graphalgo/impl/results/NormalizedCentralityResult.java
mneedham: oh you're right - they should be delegating

PR #838, opened 2019-03-15
degree centrality docs

mneedham:


mneedham:
@davidoliverSP2  we had some merge conflicts due to eigenvector. Fixed those and merging!

PR #842, opened 2019-03-20
Support {graph:'huge'} for label propagation

knutwalker:
This is not yet full huge support for LPA but allows for the huge graph to be used in case it isn't actually huge but can be used in places where a simple Graph is sufficient.
- Loading node weights and properties has been added to the huge graph
- LPA lifted its restriction to require the HeavyGraph and falls back to empty NodeWeights in case the graph does not support NodeWeights itself (GraphView should be the only one).
Since the node weight loading is added to the 'old' huge loading (without #771 ), I propose to add this first, so that I can rebase 771 and we can have the actual huge LPA later on with a smaller changeset.

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/LabelPropagationProc.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/LabelPropagationProc.java
@@ -18,11 +18,13 @@
  */
 package org.neo4j.graphalgo;
 
+import org.neo4j.graphalgo.api.Graph;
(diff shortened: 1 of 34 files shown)


mneedham:
@knutwalker is this now ready for review

knutwalker:
@mneedham yea, sure. I've added all follow-ups that aren't ready yet to another branch, so this one is independent and can be reviewed/merged.

mneedham:
@knutwalker this all looks good. Think we created some conflicts from merging the huge loading stuff earlier

PR #850, opened 2019-03-23
refactoring intro, named graphs, projected graph model

mneedham:
At a high level this PR aims to:

- [x] Simply the introduction page down to explanation, installation, basic usage
- [x] Make Cypher projection more explicit by introducing it as its own page
- [x] Explain the new Cypher projection options for dealing with duplicate relationships
- [x] Explain how to do parallel loading with Cypher projectins
- [x] Make Named graphs more explicit by introducing it as its own page and adding more explanatory text


mneedham:
@davidoliverSP2 I've done a bunch of the refactoring that we talked about around the intro page and explaining the projection stuff in more detail. Think it's now ready for you to take a look

PR #852, opened 2019-03-25
Source target ids docs

mneedham:
This PR explains the source and target ids feature that all the similarity algorithms support. I've also split up the docs for the similarity functions and procedures so that it's easier to see where each is being used

- [x] Jaccard
- [x] Cosine
- [x] Euclidean
- [x] Overlap
- [x] Pearson


mneedham:
@davidoliverSP2 I haven't finished this PR yet, but I wanted to get your advice on the restructuring I've done on the Jaccard Similarity page as I'm planning to do the same on the other pages.

I want to make the difference between functions and procedures clearer to users, but I'm not sure of whether what I've done is the best way to do that1

mneedham:
@davidoliverSP2 I've done simpler examples for the other algorithms as I'm not sure whether the first example I showed for Jaccard makes sense for them.

PR #857, opened 2019-03-26
Huge variation of LPA

knutwalker:


algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
```diff
@@ -0,0 +1,240 @@
...
+            this.direction = direction;
+            this.randomizeOrder = randomizeOrder;
+        }
+
+        @Override
+        void setExistingLabels() {
+            PrimitiveLongIterator iterator = nodes.iterator();
+            while (iterator.hasNext()) {
+                long nodeId = iterator.next();
+                long existingLabel = (long) nodeProperties.nodeWeight(nodeId, (double) nodeId);
```
jexp: could we just keep -1 or such for those nodes and then use the node-id automatically when on read its -1 ?

algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
```diff
@@ -0,0 +1,240 @@
...
+            this.nodes = RandomlySwitchingLongIterable.of(randomizeOrder, nodes);
+            this.votes = new LongDoubleScatterMap();
+        }
+
+        @Override
+        boolean computeAll() {
+            PrimitiveLongIterator iterator = nodes.iterator();
+            boolean didChange = false;
+            while (iterator.hasNext()) {
+                didChange = compute(iterator.next(), didChange);
```
jexp: we could use |= then we don't have to pass in didChange in every step

algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
```diff
@@ -0,0 +1,240 @@
...
+            return didChange;
+        }
+
+        private boolean compute(long nodeId, boolean didChange) {
+            votes.clear();
+            long partition = existingLabels.get(nodeId);
+            long previous = partition;
+            graph.forEachRelationship(nodeId, direction, this);
+            double weight = Double.NEGATIVE_INFINITY;
+            for (LongDoubleCursor vote : votes) {
```
jexp: is there a better way to compute the max, perhaps during the computation?
Feels like double work somehow.

algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
```diff
@@ -0,0 +1,240 @@
...
+                existingLabels.set(nodeId, partition);
+                return true;
+            }
+            return didChange;
+        }
+
+        @Override
+        public boolean accept(final long sourceNodeId, final long targetNodeId) {
+            long partition = existingLabels.get(targetNodeId);
+            double relationshipWeight = relationshipWeights.weightOf(sourceNodeId, targetNodeId);
```
jexp: these two are probably the most expensive operations.

in the other code for pagerank we had a cache for rel-weights and node-weights and used the variant of the consumer that got the rel-weight passed in.

jexp:
Looks good, I added a few comments that we can address later.

algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
knutwalker: The default for an empty labels array is 0, though we could set all values to -1 with a `setAll` operation more efficiently. Both 0 and -1 could also be actual user-provided values, we don't have any restriction on the partition value. If we restrict the partition to positive values, we could -1 all entries and skip this step.

algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
knutwalker: I'm not quite sure how, the votes are aggregations on the partition, we can't look for the max vote until we have gathered all votes, at least not with 100% certainty. I could employ a Heavy–Hitter/Count–min–sketch type approximation (looks like http://theory.stanford.edu/~tim/s17/l/l2.pdf could be relevant), but those have an inherent uncertainty. We might get the correct winner or might occasionally choose the wrong one and would have no way of knowing, but we can do the voting in a single pass with less memory.

algo/src/main/java/org/neo4j/graphalgo/impl/HugeLabelPropagation.java
knutwalker: `existingLabels.get` is basically an array lookup, two at the most; `relationshipWeights.weightOf` is one or two array lookups followed by a hash-map lookup; `nodeWeights.nodeWeight` is an array lookup followed by a hash-map lookup. With caching similar to what Page Rank is doing (can't do the exact same thing, as we're in huge land with long keys and direct array caches don't work anymore, requiring something like the HugeLongArray, that is backing the existingLabels, which is the closest I got to plain primitive arrays; see LongArrayBenchmark) we could get rid of the map lookups but might have to inflate the memory usage since we would have materialize all default values. On the other hand – If we have >50% of rels with a non-default weight, we actually save storage compared to the map. I would rather suggest to have a better weight mapping implementation, that would switch the internal storage from a map to an array if we have enough values and that would then be close enough to the cache so that we don't need it anymore.

jexp: Ok, makes sense, we can track this as a separate issue. Thanks!

PR #858, opened 2019-03-26
Improve randomness for LPA

knutwalker:
This one includes #857, so that one should be merged first.

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/impl/BaseLabelPropagation.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/impl/BaseLabelPropagation.java
@@ -40,7 +40,7 @@ import java.util.concurrent.ExecutorService;
 import static com.carrotsearch.hppc.Containers.DEFAULT_EXPECTED_ELEMENTS;
 import static com.carrotsearch.hppc.HashContainers.DEFAULT_LOAD_FACTOR;
 
-abstract class BaseLabelPropagation<
+public abstract class BaseLabelPropagation<
         G extends Graph,
         W extends WeightMapping,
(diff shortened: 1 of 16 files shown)


jexp:
Thanks a lot Paul.

jexp:
@knutwalker we should see in which other algorithms we could/should use those new local random generators.

How does the performance compare between LPA on Huge vs. Heavy?

knutwalker:
@jexp Louvain and Infomap could make use of the random generators.

I've benchmarked the performance of the plain iterators, but not yet of the complete Louvain impls

knutwalker:
@jexp seems like the HugeLPA doesn't do too well and the changes introduced a regression to the regular LPA as well

Screenshot_2019-04-03 JMH Visualizer

jexp:
Did  you flamegraph it?

Inlining with inheritance?

knutwalker:
I've got some here 
[lpa-flamegraphs.zip](https://github.com/neo4j-contrib/neo4j-graph-algorithms/files/3043872/lpa-flamegraphs.zip)

szarnyasg:
@knutwalker I was recently looking into LPA again, for using it in the LDBC Graphalytics benchmark. LPA docs says that:
> At every iteration of propagation, each node updates its label to the one that the maximum numbers of its neighbours belongs to. Ties are broken uniformly and randomly.
The Graphalytics benchmark requires deterministic execution, so in case of a tie, the smallest label wins.

According to the [spec](https://ldbc.github.io/ldbc_graphalytics_docs/graphalytics_spec.pdf):
image

Do you think it's possible to support such an approach to achieve determinism?

jexp:
We could add an option for that. 
Default to random option for smallest

PR #866, opened 2019-04-01
Prevent UF deadlock

### LINKED ISSUE #864: Bug: Lock in UnionFind on HeavyGraph (650M, 2.5BN)
https://community.neo4j.com/t/union-find-deadlock/6181

Hi,

I tried running the [union find algorithm](https://neo4j.com/docs/graph-algorithms/3.5/algorithms/connected-components/#algorithms-connected-components-sample-unweighted) on my graph, but it seems to have hit a deadlock. It's been running for 12+ hours but the server shows no CPU usage or I/O.

I suspect the issue is related to the size of my graph. It has 650 million nodes, 2.5 billion relationships, and is about 200 GB on disk.

My graph model is very simple: there is only one node label and one relationship label

How can I resolve this deadlock? It seems most threads are stuck here:

```
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
	at java.util.concurrent.ArrayBlockingQueue.take(ArrayBlockingQueue.java:403)
	at org.neo4j.graphalgo.impl.ParallelUnionFindQueue.lambda$compute$1(ParallelUnionFindQueue.java:85)
	at org.neo4j.graphalgo.impl.ParallelUnionFindQueue$$Lambda$450/1742932504.run(Unknown Source)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
	at org.neo4j.helpers.NamedThreadFactory$2.run(NamedThreadFactory.java:122)
```

Cypher queries used:
```
// Load graph, this completes successfully
CALL algo.graph.load('my-graph','Label','REL_TYPE',{graph:'heavy'})
  YIELD name, graph, direction, undirected, sorted, nodes, loadMillis, alreadyLoaded,
        nodeWeight, relationshipWeight, nodeProperty, loadNodes, loadRelationships;

// This query hangs
CALL algo.unionFind('User', 'FRIEND', {graph:'my-graph',write:true, partitionProperty:"partition"})
YIELD nodes, setCount, loadMillis, computeMillis, writeMillis;
```

It seems I cannot attach text files, but I have uploaded the Java stack trace and log files here:
[Stack trace](https://gist.github.com/stetra/04bb2b9ffcd027c9e1de5705c0a580aa)
[neo4j.log](https://gist.github.com/stetra/396e9f60cf9e37b49985019eacfcf14b)
[debug.log](https://gist.github.com/stetra/97d569a32bd5c78743421c12f5b81ab7)

Neo4j version: 3.5.3
Neo4j graph algorithms version: 3.5.3.3

Thanks,
Steven

knutwalker:
Fixes #864

### DIFF
--- a/algo/src/main/java/org/neo4j/graphalgo/impl/ParallelUnionFindQueue.java
+++ b/algo/src/main/java/org/neo4j/graphalgo/impl/ParallelUnionFindQueue.java
@@ -25,7 +25,11 @@ import org.neo4j.graphdb.Direction;
 
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BinaryOperator;
 import java.util.function.Function;
+
+import static org.neo4j.graphalgo.core.utils.ParallelUtil.awaitTermination;
 
 /**
  * parallel UnionFind using ExecutorService only.
@@ -45,7 +49,6 @@ public class ParallelUnionFindQueue extends GraphUnionFindAlgo<Graph, DisjointSe
     private final ExecutorService executor;
     private final int nodeCount;
     private final int batchSize;
-    private final int stepSize;
 
     public static Function<Graph, ParallelUnionFindQueue> of(ExecutorService executor, int minBatchSize, int concurrency) {
         return graph -> new ParallelUnionFindQueue(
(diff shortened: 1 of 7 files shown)


algo/src/main/java/org/neo4j/graphalgo/impl/ParallelUnionFindQueue.java
```diff
@@ -61,43 +64,64 @@
...
         }
 
-        await(futures);
+        awaitTermination(futures);
         return getStruct(queue);
     }
 
-    private void await(final List<Future<?>> futures) {
-        ParallelUtil.awaitTermination(futures);
+    static <T> void mergeTask(
```
jexp: perhaps we can generalize these in another PR later,

i.e. the expected CAS
and the actual merge operation with the push check
and move it out to a utility or (superclass minus inlining costs)

algo/src/test/java/org/neo4j/graphalgo/impl/UnionFindSafetyTest.java
```diff
@@ -0,0 +1,230 @@
...
+        public long nodeCount() {
+            return (long) nodes;
+        }
+
+        @Override
+        public void forEachRelationship(
+                final long nodeId,
+                final Direction direction,
+                final HugeRelationshipConsumer consumer) {
+            if (nodeId == 0L) {
```
jexp: why exactly at node 0 ?
i.e. at the beginning ...

jexp:
Looks good, thanks a lot.

algo/src/test/java/org/neo4j/graphalgo/impl/UnionFindSafetyTest.java
knutwalker: It's an arbitrary decision where I know that this will happen at least once. We could also always fail or randomly.

PR #869, opened 2019-04-02
Make weightProperty optional for shortest path algos

s1ck:
Co-authored-by: Max Kießling <max.kiessling@neotechnology.com>

### DIFF
--- a/tests/src/test/java/org/neo4j/graphalgo/impl/ShortestPathTest_152.java
+++ b/tests/src/test/java/org/neo4j/graphalgo/impl/ShortestPathTest_152.java
@@ -125,19 +125,18 @@ public class ShortestPathTest_152 {
         // see https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/722 and fix test afterwards
         verify(mock, times(1)).accept(eq(0.0));
         verify(mock, times(1)).accept(eq(50.0));
-        verify(mock, times(1)).accept(eq(90.0));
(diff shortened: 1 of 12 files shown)


tests/src/test/java/org/neo4j/graphalgo/impl/ShortestPathTest_152.java
```diff
@@ -125,19 +125,18 @@ public void testDirectRoadOrRail() {
         // see https://github.com/neo4j-contrib/neo4j-graph-algorithms/issues/722 and fix test afterwards
         verify(mock, times(1)).accept(eq(0.0));
         verify(mock, times(1)).accept(eq(50.0));
-        verify(mock, times(1)).accept(eq(90.0));
-        verify(mock, times(1)).accept(eq(120.0));
-        verify(mock, times(1)).accept(eq(140.0));
+        verify(mock, times(1)).accept(eq(80.0));
```
s1ck: @knutwalker we have *NO* idea why this commit changed the expected result to the correct one. Maybe you can explain?

knutwalker: @s1ck beats me, the test in question doesn't even call the procedure… Do you maybe have a newer version of 3.5? IIRC, the test correctness depends on whether we get the correct relationship first from the kernel. Changes there could change the outcome of this test.

PR #880, opened 2019-04-09
Degree cutoff skip values

mneedham:


mneedham:
@jexp for this one, at the moment we are checking the degreeCutoff against the size of each weights list, which doesn't make sense as many of those values could be skip values! This PR compares degreeCutoff against the size of the weight list excluding items that should be skipped

PR #895, opened 2019-04-17
Amend the image of the Graph Algos book in the docs

davidoliverSP2:
We have a limitation in our Asciidoc to XML conversion, where Docbook doesn't bring forward the link.

I've amended the way we present this, so users have a link to click.


mneedham:
Looks good to me.

PR #904, opened 2019-06-26
Replace instance of algo.getNodeById with algo.asNode

davidoliverSP2:


Mats-SX:
This PR has been moved to the new source location for graph algo docs.

PR #907, opened 2019-07-18
Fixed calculation of denominator for Jaccard-similarity

oschlueter:
The current implementation of Jaccard-similarity doesn't discard dupliate input values when calculating the denominator. I identified this issue by calculating Jaccard on identical input containing duplicates which didn't return `1.0` for which I added test cases.


oschlueter:
When creating the test cases I forgot to set the call to `Similarities::overlapSimilarity`. Once fixed I saw that the calculation is also affected so I added a propsed fix for overlapSimilarity as well.

